PR #9, opened 2015-01-15
Solution to unresponsive repl in Windows OS

mavbozo:
in Windows OS, the repl is not responsive. for example, typing `(+ 1 1)` and `ENTER` does not return anything. The cursor just drops to new line.

inf-clojure-unresponsive-repl-windows-os

I found the solution here https://groups.google.com/forum/#!topic/leiningen/48M-xvcI2Ng.

Basically, just create file `.jline.rc` and put this line on this file.

`jline.terminal=unsupported`

I don't know where to put it. So, I decided just to add Troubleshooting section in README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,19 @@ Add the following to your Emacs config to enable
 
 `M-x inf-clojure` or `C-c C-z` within a Clojure source file.
 
+## Troubleshooting
+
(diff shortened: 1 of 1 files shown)


bbatsov:
:+1: Thanks!

PR #10, opened 2015-01-16
Configuration option for disabling completions in minor mode

### LINKED ISSUE #5: Make it possible to turn off completions in clojure-mode buffers
If the REPL is "busy" inf-clojure will interrupt text entry when editing text in clojure-mode buffers. Some users might not mind, some do. Would be good to be able to disable this via configuration.

Please separate this config from disabling completions in the `*inf-clojure*` buffer, which you probably always want to have turned on.

martintrojer:
fixes #5

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -61,6 +61,11 @@ mode.  Default is whitespace followed by 0 or 1 single-letter colon-keyword
   :type 'regexp
   :group 'inf-clojure)
 
+(defcustom inf-clojure-minor-mode-completions t
+  "If non-nil, completions will be used in inf-clojure minor mode."
+  :type 'boolean
+  :group 'inf-clojure)
+
 (defvar inf-clojure-mode-map
   (let ((map (copy-keymap comint-mode-map)))
     (define-key map "\C-x\C-e" 'inf-clojure-eval-last-sexp)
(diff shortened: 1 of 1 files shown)


inf-clojure.el
```diff
@@ -126,8 +131,9 @@ The following commands are available:
 \\{inf-clojure-minor-mode-map}"
   :lighter "" :keymap inf-clojure-minor-mode-map
   (make-local-variable 'completion-at-point-functions)
-  (add-to-list 'completion-at-point-functions
-               #'inf-clojure-completion-at-point))
+  (when inf-clojure-minor-mode-completions
```
bbatsov: This is not a good solution as it will require the users to set this **before** loading inf-clojure, which will certainly lead to confusion. Generally such config option should be consulted at runtime to avoid such problems. This can also be done much simpler by setting `completion-at-point-functions` to `nil` in the mode's hook.

martintrojer:
Thanks for the feedback. 

I'd be interested to know how the 'setting before loading' problem is different for the other configuration variables like inf-clojure-prompt-read-only.

bbatsov:
> I'd be interested to know how the 'setting before loading' problem is different for the other configuration variables like inf-clojure-prompt-read-only.

It is, but it's pretty hard to sidestep it there and I'm assuming pretty much no one will ever want a non-read-only prompt (that's an option I might end up removing some day).

martintrojer:
I have it set to non-read-only since I constantly want to clear the REPL buffer! :-)

Now I just have to figure out how to set the minor modes (local) `completion-at-point-functions` to nil in the mode hook. Can that be done without any code changes to inf-clojure code?

bbatsov:
> I have it set to non-read-only since I constantly want to clear the REPL buffer! :-)

We should simply add a command for this. The command temporary make remove inhibit the read-only config, clear the buffer, and restore it.

> Now I just have to figure out how to set the minor mode's (local) completion-at-point-functions to nil in the mode hook. Can that be done without any code changes to inf-clojure code?

Invoking `(setq completion-at-point-functions nil)` in a hook function should be enough.

bbatsov:
I've added a command for clearing the REPL buffer. It's bound to `C-c M-o` (as in CIDER).

martintrojer:
Ok thanks, closing this PR

Please keep in mind no to block in the minor mode for the eldoc stuff aswell.

PR #12, opened 2015-04-30
Add new-line mark after inf-clojure-arglist-command

ayato0211:
I tried to use `C-c C-a` as the inf-clojure-show-arglist command, then I couldn't see any output.
I'm not sure, but maybe this command needed new-line mark after that command.
Thanks.


bbatsov:
Have you tested whether this change solves your problem?

ayato0211:
Yes. It seems fine to me.

bbatsov:
:+1: Thanks!

PR #15, opened 2015-09-19
inf-clojure capf completion function require inf-clojure inferior buffer opened

stardiviner:


bbatsov:
I don't think this is a good approach (inserting a hook like this). Probably it'd be better to either complain there's no `inf-clojure` running or turn some commands into no-ops.

stardiviner:
@bb

bbatsov:
Yeah, that'd be fine.

PR #19, opened 2015-10-11
Too many newlines

### LINKED ISSUE #18: inf-clojure-eval-region always sends a newline to inf-clojure process, even when not needed
Currently `inf-clojure-eval-region` always sends a newline after sending the region text.

This often leads to showing multiple prompts because in most cases the region already contains a newline at the end. Then you end up with something like this:

```clojure
{:a 1}
user=>
user=>
```

Would be nice if it only added a newline when it doesn't have one already. Can create a PR if you like.

otijhuis:
Modify `inf-clojure-eval-region` as follows:

- if region already ends with one or more newlines replace it with a single newline
- if region doesn't end with a newline then add one

resolves #18

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -323,8 +323,12 @@ of `inf-clojure-program').  Runs the hooks from
   "Send the current region to the inferior Clojure process.
 Prefix argument AND-GO means switch to the Clojure buffer afterwards."
   (interactive "r\nP")
-  (comint-send-region (inf-clojure-proc) start end)
-  (comint-send-string (inf-clojure-proc) "\n")
+  ;; replace multiple newlines at the end of the region by a single one
+  ;; or add one if there was no newline
+  (let ((str (replace-regexp-in-string
+              "[\n]*\\'" "\n"
+              (buffer-substring-no-properties start end))))
+    (comint-send-string (inf-clojure-proc) str))
   (if and-go (inf-clojure-switch-to-repl t)))
 
 (defun inf-clojure-eval-string (code)


inf-clojure.el
```diff
@@ -323,8 +323,12 @@ of `inf-clojure-program').  Runs the hooks from
   "Send the current region to the inferior Clojure process.
 Prefix argument AND-GO means switch to the Clojure buffer afterwards."
   (interactive "r\nP")
-  (comint-send-region (inf-clojure-proc) start end)
-  (comint-send-string (inf-clojure-proc) "\n")
+  ;; replace multiple newlines at the end of the region by a single one
+  ;; or add one if there was no newline
+  (let ((str (replace-regexp-in-string
+              "[\n]*\\'" "\n"
```
bbatsov: I'm not sure we don't have account for Windows here. There a newline is `\r\n`.

otijhuis:
Hmm, totally forgot about Windows.

Have to try it out. Otherwise this regex should probably work: `"\\(\r\n?\\|\n\\)*\\'"`

Have a windows virtual machine, not sure it has Emacs and such installed already.

bbatsov:
Well, let's go with this for now. It's still an improvement and I guess Windows users will complain if they run into into trouble. :-)

otijhuis:
Ok :). I'm trying it on windows in a bit. If there are any issues I'll update it and create another PR.

otijhuis:
Looks like it works on windows as well. Even converted a clj file to have windows line endings.
No extra newlines in the output.

bbatsov:
Great.

PR #21, opened 2015-10-14
Refactoring

syohex:
- Pass and-go parameter to inf-clojure-eval-region
- Use point-min and point-max instead of beginning-of-buffer and end-of-buffer

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -352,19 +352,16 @@ Prefix argument AND-GO means switch to the Clojure buffer afterwards."
     (end-of-defun)
     (let ((end (point)) (case-fold-search t))
       (beginning-of-defun)
-      (inf-clojure-eval-region (point) end)))
-  (if and-go (inf-clojure-switch-to-repl t)))
+      (inf-clojure-eval-region (point) end and-go))))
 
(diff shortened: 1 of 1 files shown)


inf-clojure.el
```diff
@@ -352,19 +352,14 @@ Prefix argument AND-GO means switch to the Clojure buffer afterwards."
...
 Prefix argument AND-GO means switch to the Clojure buffer afterwards."
   (interactive "P")
-  (save-excursion
-    (end-of-buffer)
-    (let ((end (point)) (case-fold-search t))
-      (beginning-of-buffer)
-      (inf-clojure-eval-region (point) end)))
-  (if and-go (inf-clojure-switch-to-repl t)))
+  (let ((case-fold-search t))
+    (inf-clojure-eval-region (point-min) (point-max) and-go)))
```
bbatsov: I was wondering about this - `point-mix` and `point-max` are affected by narrowing, which might a feature or a bug or the user's point of view.

syohex:
Updated for narrowing region.

bbatsov:
Guess this will do. Thanks!

PR #22, opened 2015-10-27
add eldoc support

otijhuis:
This probably needs some cleaning up and maybe some changes.
Would be really nice if you could take a look.

I started out with the code from cider and changed things where needed. Haven't hit any performance issues so far. Eldoc works in clojure buffer and repl.

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -40,6 +40,8 @@
 
 (require 'comint)
 (require 'clojure-mode)
+(require 'eldoc)
+(require 'thingatpt)
 
 
 (defgroup inf-clojure nil
@@ -131,6 +133,7 @@ The following commands are available:
 
 \\{inf-clojure-minor-mode-map}"
   :lighter "" :keymap inf-clojure-minor-mode-map
+  (inf-clojure-eldoc-setup)
   (make-local-variable 'completion-at-point-functions)
   (add-to-list 'completion-at-point-functions
                #'inf-clojure-completion-at-point))
@@ -247,6 +250,7 @@ to continue it."
   (setq comint-prompt-regexp inf-clojure-prompt)
   (setq mode-line-process '(":%s"))
   (clojure-mode-variables)
+  (inf-clojure-eldoc-setup)
   (setq comint-get-old-input #'inf-clojure-get-old-input)
   (setq comint-input-filter #'inf-clojure-input-filter)
   (set (make-local-variable 'comint-prompt-read-only) inf-clojure-prompt-read-only)
@@ -429,10 +433,14 @@ Used by this command to determine defaults."
                                         ; doesn't need an exact name
   (comint-check-source file-name) ; Check to see if buffer needs saved.
   (setq inf-clojure-prev-l/c-dir/file (cons (file-name-directory    file-name)
-                                        (file-name-nondirectory file-name)))
+                                            (file-name-nondirectory file-name)))
   (comint-send-string (inf-clojure-proc)
                       (format inf-clojure-load-command file-name))
   (inf-clojure-switch-to-repl t))
+
+(defun inf-clojure-connected-p ()
+  "Return t if inferior Clojure is currently connected, nil otherwise."
+  (not (null inf-clojure-buffer)))
 
 
 ;;; Documentation functions: function doc, var doc, arglist, and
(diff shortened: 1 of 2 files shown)


bbatsov:
I'd suggest killing the separate files and just putting everything in `inf-clojure.el`. The package is way simpler than CIDER, so breaking it up doesn't add much value. I'd also suggest against the backports from `subr-x`, as the codebase is tiny and we don't really need them that much.

bbatsov:
P.S. Capitalize the commit messages.

otijhuis:
Sure thing. I just kept the same structure for now so it's easier for me to find things in Cider.
Thanks for the feedback :). I'll fix those things soon.

Currently trying to get it working for ClojureScript, which is a bit of a pain it seems.

bbatsov:
OK, take your time. We're definitely not in a hurry.

otijhuis:
I cleaned things up and added everything that was needed to `inf-clojure.el`. Not enabling eldoc mode when the major mode is ClojureScript currently. Still looking for a solution. Asked Bruce Hauman for some help as well but he didn't see an easy solution either.

Because of ClojureScript I haven't enabled eldoc mode in the REPL by default.

You can run
```elisp
(inf-clojure-eldoc-setup)
(eldoc-mode +1)
```
when the `inf-clojure` buffer is active and then it works fine in the REPL as well. For Clojure that is.

bbatsov:
Guess you can add this info the README, so it'd be easier for users to find it.

otijhuis:
Sure thing. I'll add it. Or is there an easy way to find out if a cljs or clj REPL is running?

otijhuis:
Just noticed that I can always do the `(inf-clojure-eldoc-setup)` anyway. Saves a step. I'll update the PR.

otijhuis:
Only `M-x eldoc-mode` needed now in `*inf-clojure*` buffer. Updated README.

inf-clojure.el
```diff
@@ -131,6 +134,10 @@ The following commands are available:
 
 \\{inf-clojure-minor-mode-map}"
   :lighter "" :keymap inf-clojure-minor-mode-map
+  (inf-clojure-eldoc-setup)
+  ;; No ElDoc support for ClojureScript yet
+  (unless (equal major-mode 'clojurescript-mode)
```
bbatsov: It's generally better to use `derived-mode-p`, as it will work with derived modes as well.

bbatsov:
Btw, I now remembered we don't turn on `eldoc-mode` by default in CIDER, so we should probably not turn it on by default here as well. We just advise people there to manually add it to their mode hooks.

Btw, I'm guessing the eldoc-code could check that's in a Clojure REPL, so that's it's safe to have it turned on in a ClojureScript REPL.

otijhuis:
Ok, no problem. I'll change it to off by default then.

Been thinking about doing the check in the eldoc-code but not sure what the best, and performant, way is to check if it's a Clojure REPL.

bbatsov:
Guess you can simply check about the existence of some cljs ns or dynamic var. That should be quite fast.

otijhuis:
Updated the PR:

- eldoc is no longer enabled by default
- deleted a debug message I forgot to remove
- updated the README on how to enable eldoc

In ClojureScript the debug message was still showing before, this made it annoying.
Tried it just now with ClojureScript. No problems, just won't show anything in the echo area.
So you should be able to leave ElDoc enabled, even in cljs buffers/REPLs. Haven't seen
any troublesome performance issues so far.

bbatsov:
Apart from my small remarks, things are looking good. We're pretty close to merging this.

otijhuis:
Sorry about the `cl-lib`. Copied a lot from `cider-eldoc.el` but turned out I didn't need certain functions. Those used `cl-lib`.

otijhuis:
Already removed it.

otijhuis:
Did you have any other remarks besides `cl-lib`? Force pushed some stuff to keep it clean. Guess that removed the comments as well.

Btw, what is your regular workflow with PR's? Do I need to force push or fix things in new commits?
Still figuring out the best way to do things. If you have any advice let me know. Want to make it as easy for you as possible and I'm still rather new at this PR stuff :)

bbatsov:
> Did you have any other remarks besides cl-lib? Force pushed some stuff to keep it clean. Guess that removed the comments as well.

They were about the text in the README. Guess you can see the email notification or check github's notifications centre. 

> Btw, what is your regular workflow with PR's? Do I need to force push or fix things in new commits?
> Still figuring out the best way to do things. If you have any advice let me know. Want to make it as easy for you as possible and I'm still rather new at this PR stuff :)

I also address comments by force-pushing. Doing new commits is fine as well, as long as they get properly squashed before being merged in the end.

otijhuis:
Ok, thanks for the info. Updated the README.

PR #23, opened 2015-11-09
Fix for older Emacs

syohex:
Because 'if-let' is defined only on development Emacs(version 25).


bbatsov:
:+1: Yeah, we've inlined the definition in cider and I totally forgot about this.

PR #27, opened 2015-12-15
Switch to repl is optional when load file

oneness:
This commit adds current-prefix-arg check before switch to repl. So one
can use C-u M-x inf-clojure-load-file to switch to repl

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -436,7 +436,8 @@ Used by this command to determine defaults."
                                             (file-name-nondirectory file-name)))
   (comint-send-string (inf-clojure-proc)
                       (format inf-clojure-load-command file-name))
-  (inf-clojure-switch-to-repl t))
+  (when current-prefix-arg
+    (inf-clojure-switch-to-repl t)))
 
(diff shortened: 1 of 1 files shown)


inf-clojure.el
```diff
@@ -436,7 +436,8 @@ Used by this command to determine defaults."
                                             (file-name-nondirectory file-name)))
   (comint-send-string (inf-clojure-proc)
                       (format inf-clojure-load-command file-name))
-  (inf-clojure-switch-to-repl t))
+  (when current-prefix-arg
```
bbatsov: It'd be better to convert this to a properly `(interactive "P")` argument. The usage of `current-prefix-arg` is quite uncommon. Guess `file-name` will have to be made optional as well, for this to work.

oneness: Sure, will make a PR soon.

oneness: I am using inf-clojure-eval-buffer which has what I wanted. So I do not see the need of modifying inf-clojure-load-file now.  So you can close this as non issue. 

Thanks.

PR #32, opened 2016-02-01
Add possibility to customize more vars

slipset:
Planck, https://github.com/mfikes/planck, the Clojurscript
repl on OSX, uses somewhat different commands to achieve
things like set-ns, apropos, dir etc. This commit
lets the user customize these commands.

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -450,45 +450,63 @@ Used by this command to determine defaults."
 ;;; Command strings
 ;;; ===============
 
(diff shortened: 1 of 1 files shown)


bbatsov:
:+1: No argument from me. I wonder why I didn't made them defcustoms from day 1...

bbatsov:
Btw, now that #44 is in, would you be interested in added a planck REPL type and it's relevant forms for `set-ns`, `apropos` and so on?

PR #34, opened 2016-06-20
Accept cons pair for tcp connections.

mahinshaw:
Coming takes a cons pair for connecting to tcp connections. This allows `inf-clojure` (or setting `inf-clojure-program`) to accept a cons pair for connecting to a clojure socket repl server. I tested this with planck via `planck -n 5555`.

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -139,8 +139,11 @@ The following commands are available:
                #'inf-clojure-completion-at-point))
 
 (defcustom inf-clojure-program "lein repl"
-  "Program name for invoking an inferior Clojure in Inferior Clojure mode."
-  :type 'string
+  "Either a program name or a connection cons pair consisting of a host
+and port number (e.g. (\"localhost\" . 5555)), for invoking an inferior Clojure
+in Inferior Clojure mode."
+  :type '(choice (string)
+                 (cons string integer))
   :group 'inf-clojure)
 
 (defcustom inf-clojure-load-command "(clojure.core/load-file \"%s\")\n"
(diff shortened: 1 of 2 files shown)


inf-clojure.el
```diff
@@ -322,7 +322,9 @@ of `inf-clojure-program').  Runs the hooks from
   (if (not (comint-check-proc "*inf-clojure*"))
       ;; run the new process in the project's root when in a project folder
       (let ((default-directory (inf-clojure-project-root))
-            (cmdlist (split-string cmd)))
+            (cmdlist (if (consp cmd)
```
bbatsov: I'm guessing the related defcustom should be updated as well.

mahinshaw:
Sorry about that. I updated the defcustom.

bbatsov:
Btw, what exactly is the reasoning behind this change? I'm guessing that `split-string` won't work properly in such scenarios? 

I'm not 100% certain what are you trying to achieve here.

mahinshaw:
I was trying to connect to a clojure socket repl. And since a cons pair for tcp connections is part of the comint API, it made sense to leverage that option.

bbatsov:
Got it. Well, I guess you should make this a bit more explicit in the `defcustom` and maybe mention it in the README. I didn't know about this `comint` functionality.

mahinshaw:
@bbatsov I updated the defcustom doc string, and the readme. Let me know if that description is what you are looking for.

bbatsov:
👍

bbatsov:
Btw, see #44 - I think we can extend this to something that would offer a much better planck support out-of-the-box

PR #44, opened 2017-03-01
Support REPL flavors and Lumo

arichiardi:
This patch adds the `inf-clojure-repl-flavor` defcustom in order to
support different kind of configurations based on the selected flavor.
The default flavor is `'clojure`, and the other supported one is `'lumo`.
It also adds defcustoms to further customize hostname, port and
classpath used in the Lumo flavor.

--

The defcustoms for the following are already ok:
 * `inf-clojure-set-ns-form`
 * `inf-clojure-macroexpand-form`
 * `inf-clojure-macroexpand-1-form`

TODOs
 * `(setq inf-clojure-var-source-form "(lumo.repl/source %s)")` - https://github.com/anmonteiro/lumo/issues/84
 * `(setq inf-clojure-ns-vars-form "(lumo.repl/dir %s)")` - https://github.com/anmonteiro/lumo/issues/87
 * `(setq inf-clojure-var-apropos-form "(lumo.repl/apropos %s)")` - https://github.com/anmonteiro/lumo/issues/86
 * `(setq inf-clojure-arglist-form "(lumo.repl/get-arglists \"%s\")")` - https://github.com/anmonteiro/lumo/pull/88

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -158,7 +158,7 @@ The following commands are available:
 (defcustom inf-clojure-lein-cmd "lein repl"
   "The command used to start a Clojure REPL for Leiningen projects.
 
-Alternative you can specify a TCP connection cons pair, instead
+Alternatively you can specify a TCP connection cons pair, instead
 of command, consisting of a host and port
 number (e.g. (\"localhost\" . 5555)).  That's useful if you're
 often connecting to a remote REPL process."
@@ -168,7 +168,7 @@ often connecting to a remote REPL process."
 (defcustom inf-clojure-boot-cmd "boot repl"
   "The command used to start a Clojure REPL for Boot projects.
 
-Alternative you can specify a TCP connection cons pair, instead
+Alternatively you can specify a TCP connection cons pair, instead
 of command, consisting of a host and port
 number (e.g. (\"localhost\" . 5555)).  That's useful if you're
 often connecting to a remote REPL process."
@@ -178,7 +178,7 @@ often connecting to a remote REPL process."
 (defcustom inf-clojure-generic-cmd "lein repl"
   "The command used to start a Clojure REPL outside Lein/Boot projects.
 
-Alternative you can specify a TCP connection cons pair, instead
+Alternatively you can specify a TCP connection cons pair, instead
 of command, consisting of a host and port
 number (e.g. (\"localhost\" . 5555)).  That's useful if you're
 often connecting to a remote REPL process."
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -44,6 +44,7 @@
 (require 'thingatpt)
 (require 'ansi-color)
 (require 'cl-lib)
+(require 'subr-x)
 
 
 (defgroup inf-clojure nil
@@ -148,6 +149,7 @@ The following commands are available:
 
 \\{inf-clojure-minor-mode-map}"
   :lighter "" :keymap inf-clojure-minor-mode-map
+  (setq comint-input-sender 'inf-clojure--send-string)
   (inf-clojure-eldoc-setup)
   (make-local-variable 'completion-at-point-functions)
   (add-to-list 'completion-at-point-functions
@@ -182,6 +184,30 @@ number (e.g. (\"localhost\" . 5555)).  That's useful if you're
 often connecting to a remote REPL process."
   :type '(choice (string)
                  (cons string integer)))
+
+(defvar-local inf-clojure-repl-type nil
+  "Symbol to define your REPL type.
+Its root binding is nil and it can be further customized using
+either `setq-local` or an entry in `.dir-locals.el`." )
+
+(defun inf-clojure--detect-type (proc)
+  "Identifies the current REPL type for PROC."
+  (cond
+   ((inf-clojure--lumo-p proc) 'lumo)
+   (t 'clojure)))
+
+(defun inf-clojure--set-repl-type (proc)
+  "Set the REPL type if has not already been set."
+  (when (not inf-clojure-repl-type)
+    (setq inf-clojure-repl-type (inf-clojure--detect-type proc))))
+
+(defun inf-clojure--send-string (proc string)
+  "A custom `comint-input-sender` / `comint-send-string`.
+Perform the required side effects on every send for PROC and
+STRING (for example set the buffer local REPL type).  It should
+be used instead of `comint-send-string`."
+  (inf-clojure--set-repl-type proc)
+  (comint-simple-send proc string))
 
 (defcustom inf-clojure-load-command "(clojure.core/load-file \"%s\")\n"
   "Format-string for building a Clojure expression to load a file.
@@ -296,6 +322,7 @@ If `comint-use-prompt-regexp' is nil (the default), \\[comint-insert-input] on o
 Paragraphs are separated only by blank lines.  Semicolons start comments.
 If you accidentally suspend your process, use \\[comint-continue-subjob]
 to continue it."
+  (setq comint-input-sender 'inf-clojure--send-string)
   (setq comint-prompt-regexp inf-clojure-comint-prompt-regexp)
   (setq mode-line-process '(":%s"))
   (clojure-mode-variables)
@@ -408,12 +435,12 @@ Prefix argument AND-GO means switch to the Clojure buffer afterwards."
   (let ((str (replace-regexp-in-string
               "[\n]*\\'" "\n"
               (buffer-substring-no-properties start end))))
-    (comint-send-string (inf-clojure-proc) str))
+    (inf-clojure--send-string (inf-clojure-proc) str))
   (if and-go (inf-clojure-switch-to-repl t)))
 
 (defun inf-clojure-eval-string (code)
   "Send the string CODE to the inferior Clojure process to be executed."
-  (comint-send-string (inf-clojure-proc) (concat code "\n")))
+  (inf-clojure--send-string (inf-clojure-proc) (concat code "\n")))
 
 (defun inf-clojure-eval-defun (&optional and-go)
   "Send the current defun to the inferior Clojure process.
@@ -504,8 +531,8 @@ The prefix argument SWITCH-TO-REPL controls whether to switch to REPL after the 
     (comint-check-source file-name) ; Check to see if buffer needs saved.
     (setq inf-clojure-prev-l/c-dir/file (cons (file-name-directory    file-name)
                                               (file-name-nondirectory file-name)))
-    (comint-send-string (inf-clojure-proc)
-                        (format inf-clojure-load-command file-name))
+    (inf-clojure--send-string (inf-clojure-proc)
+                              (format inf-clojure-load-command file-name))
     (when switch-to-repl
       (inf-clojure-switch-to-repl t))))
 
@@ -521,10 +548,23 @@ The prefix argument SWITCH-TO-REPL controls whether to switch to REPL after the 
 ;;; Command strings
 ;;; ===============
 
-(defcustom inf-clojure-var-doc-command
+(defcustom inf-clojure-var-doc-form
   "(clojure.repl/doc %s)\n"
   "Command to query inferior Clojure for a var's documentation."
   :type 'string)
+
+(defcustom inf-clojure-var-doc-form-lumo
+  "(lumo.repl/doc %s)\n"
+  "Lumo command to query inferior Clojure for a var's documentation."
+  :type 'string)
+
+(defun inf-clojure-var-doc-form ()
+  "Return the form to query inferior Clojure for a var's documentation.
+If you are using REPL types, it will pickup the most approapriate
+`inf-clojure-var-doc-form` variant."
+  (pcase inf-clojure-repl-type
+    (lumo inf-clojure-var-doc-form-lumo)
+    (_ inf-clojure-var-doc-form)))
 
 (defcustom inf-clojure-var-source-command
   "(clojure.repl/source %s)\n"
@@ -541,10 +581,23 @@ The prefix argument SWITCH-TO-REPL controls whether to switch to REPL after the 
   "Command to query inferior Clojure for a function's arglist."
   :type 'string)
 
-(defcustom inf-clojure-completion-command
+(defcustom inf-clojure-completion-form
   "(complete.core/completions \"%s\")\n"
   "Command to query inferior Clojure for completion candidates."
   :type 'string)
+
+(defcustom inf-clojure-completion-form-lumo
+  "(doall (map str (lumo.repl/get-completions \"%s\")))\n"
+  "Lumo form to query inferior Clojure for completion candidates."
+  :type 'string)
+
+(defun inf-clojure-completion-form ()
+  "Return the form to query inferior Clojure for a var's documentation.
+If you are using REPL types, it will pickup the most approapriate
+`inf-clojure-completion-form` variant."
+  (pcase inf-clojure-repl-type
+    (lumo inf-clojure-completion-form-lumo)
+    (_ inf-clojure-completion-form)))
 
 (defcustom inf-clojure-ns-vars-command
   "(clojure.repl/dir %s)\n"
@@ -618,9 +671,9 @@ The value is nil if it can't find one."
 
 (defun inf-clojure-show-var-documentation (var)
   "Send a command to the inferior Clojure to give documentation for VAR.
-See variable `inf-clojure-var-doc-command'."
+See function `inf-clojure-var-doc-form'."
   (interactive (inf-clojure-symprompt "Var doc" (inf-clojure-var-at-pt)))
-  (comint-proc-query (inf-clojure-proc) (format inf-clojure-var-doc-command var)))
+  (comint-proc-query (inf-clojure-proc) (format (inf-clojure-var-doc-form) var)))
 
 (defun inf-clojure-show-var-source (var)
   "Send a command to the inferior Clojure to give source for VAR.
@@ -683,7 +736,7 @@ See variable `inf-clojure-macroexpand-command'.
 With a prefix arg MACRO-1 uses `inf-clojure-macroexpand-1-command'."
   (interactive "P")
   (let ((last-sexp (buffer-substring-no-properties (save-excursion (backward-sexp) (point)) (point))))
-    (comint-send-string
+    (inf-clojure--send-string
      (inf-clojure-proc)
      (format (if macro-1
                  inf-clojure-macroexpand-1-command
@@ -710,7 +763,7 @@ See variable `inf-clojure-buffer'."
     (unwind-protect
         (let ((completion-snippet
                (format
-                inf-clojure-completion-command (substring-no-properties expr))))
+                (inf-clojure-completion-form) (substring-no-properties expr))))
           (process-send-string proc completion-snippet)
           (while (and (not (string-match inf-clojure-prompt kept))
                       (accept-process-output proc 2)))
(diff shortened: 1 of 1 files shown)


inf-clojure.el
```diff
@@ -770,6 +809,77 @@ Return the number of nested sexp the point was over or after."
   (setq-local eldoc-documentation-function #'inf-clojure-eldoc)
   (apply #'eldoc-add-command inf-clojure-extra-eldoc-commands))
 
-(provide 'inf-clojure)
+;;;; Lumo
+;;;; ====
 
+(defgroup lumo nil
```
arichiardi: Probably I don't need this one, I will delete it.

inf-clojure.el
```diff
@@ -770,6 +809,77 @@ Return the number of nested sexp the point was over or after."
...
+  "Run an external Lumo process (REPL) in an Emacs buffer."
+  :group 'inf-clojure)
+
+;; AR - TODO Alternatively you can specify a command string that will be called,
+;; which should return a string.
+(defcustom inf-clojure-lumo-classpath-generator "cp"
```
arichiardi: This name looks like it is arbitrary, but it is kind of standard to call it something like this.

inf-clojure.el
```diff
@@ -770,6 +809,77 @@ Return the number of nested sexp the point was over or after."
...
+
+(defcustom inf-clojure-lumo-hostname "localhost"
+  "The hostname used to launch lumo.
+Note that only if port is specified this will be used."
+  :type '(choice (string))
+  :group 'lumo)
+
+(defcustom inf-clojure-lumo-port ""
+  "The port used to launch lumo.
+Note that if port is specified, inf-clojure will connect to it by
```
arichiardi: Will rewrite this as well.

inf-clojure.el
```diff
@@ -770,6 +824,61 @@ Return the number of nested sexp the point was over or after."
...
+(defcustom inf-clojure-lumo-port ""
+  "The port used to launch lumo.
+Note that if port is specified, inf-clojure will connect to the
+REPL using HTTP socket.  No process will be actually started."
+  :type '(choice string integer)
+  :group 'lumo)
+
+;; AR - TODO Alternatively you can specify a command string that will be called,
+;; which should return a string.
+(defcustom inf-clojure-lumo-classpath-generator "cp"
```
arichiardi: This approach could be extended to all REPL flavors, the all need a classpath to be built if using the "command" approach. Socket of course will already have everything setup. Lumo does not differ in this aspect so we could even get rid of all this. I personally settled on the socket connection but it is handy sometimes to start things in place.

inf-clojure.el
```diff
@@ -770,6 +824,61 @@ Return the number of nested sexp the point was over or after."
...
+  :group 'lumo)
+
+;; AR - not used but left here because it is a possible sanity check
+(defun inf-clojure--lumo-mode-p ()
+  "Return true if the lumo is the target REPL."
+  (comint-send-string (inf-clojure-proc) "(js/global.hasOwnProperty \"$$LUMO_GLOBALS\")"))
+
+(defun inf-clojure--lumo-program ()
+  "Return inf-clojure-program for lumo."
+  (if (inf-clojure--empty-string-p inf-clojure-lumo-port)
```
arichiardi: an explicit `program` is returned only if the `port` is empty.

inf-clojure.el
```diff
@@ -559,6 +575,26 @@ The value is nil if it can't find one."
   "Return the name of the symbol at point, otherwise nil."
   (or (thing-at-point 'symbol) ""))
 
+(defun inf-clojure--empty-string-p (string)
```
bbatsov: Isn't there such a function in `subr-x.el`?

inf-clojure.el
```diff
@@ -208,6 +209,21 @@ whichever process buffer you want to use.")
 
 (put 'inf-clojure-mode 'mode-class 'special)
 
+(defcustom inf-clojure-repl-flavor 'clojure
```
bbatsov: I really think there should be only a default REPL type and some command that let's easily pick the type of REPL you want to create. Otherwise creating both a lumo repl and a clojure repl would be really frustrating.

inf-clojure.el
```diff
@@ -770,6 +806,64 @@ Return the number of nested sexp the point was over or after."
...
+  ;; https://github.com/anmonteiro/lumo/issues/84
+  ;; (setq inf-clojure-var-source-command "(lumo.repl/source %s)")
+  ;; https://github.com/anmonteiro/lumo/issues/87
+  ;; (setq inf-clojure-ns-vars-command "(lumo.repl/dir %s)")
+  ;; https://github.com/anmonteiro/lumo/issues/86
+  ;; (setq inf-clojure-var-apropos-command "(lumo.repl/apropos %s)")
+
+  ;; Uncomment after https://github.com/anmonteiro/lumo/pull/88
+  ;; (setq inf-clojure-arglist-command "(lumo.repl/get-arglists \"%s\")")
+  (setq inf-clojure-var-doc-command "(lumo.repl/doc %s)")
```
bbatsov: Modifying defcustoms like this is a bad idea in general. We should make those variables `buffer-local` so that changing them for one REPL won't affect the rest.

inf-clojure.el
```diff
@@ -770,6 +806,64 @@ Return the number of nested sexp the point was over or after."
...
+files."
+  :type 'string
+  :group 'lumo)
+
+;; AR - not used but left here because it is a possible sanity check
+(defun inf-clojure--lumo-mode-p ()
+  "Return true if the lumo is the target REPL."
+  (comint-send-string (inf-clojure-proc) "(js/global.hasOwnProperty \"$$LUMO_GLOBALS\")"))
+
+(defun inf-clojure--lumo-program ()
```
bbatsov: I think this name is a bit confusing. This doesn't return the lumo program, but some command complete with parameters.

inf-clojure.el
```diff
@@ -559,6 +575,26 @@ The value is nil if it can't find one."
...
+  "Read the classpath from the input CLASSPATH-OPTION."
+  (let ((classpath-file-path (concat (inf-clojure-project-root) classpath-option)))
+    (cond
+     ((and (file-exists-p classpath-file-path) (file-readable-p classpath-file-path))
+      (f-read classpath-file-path))
+     ;; TODO launch a command that returns the classpath string?
+     ((message (concat "Option \"" classpath-option "\" was not a (readable) file, the classpath will be empty."))
+      ""))))
+
+(defun inf-clojure--maybe-set-program (program)
```
bbatsov: I don't quite get the point of this.

inf-clojure.el
```diff
@@ -208,6 +209,21 @@ whichever process buffer you want to use.")
...
+The default flavor is 'clojure, 'lumo is the other supported
+one."
+  :type 'symbol
+  :options '(clojure lumo)
+  :group 'inf-clojure)
+
+(defun inf-clojure--flavor-setup ()
+  "Setup inf-clojure defcustoms depending on the choose flavor."
+  (pcase inf-clojure-repl-flavor
+    (lumo (progn (message "[inf-clojure] will switch to the Lumo flavor")
```
bbatsov: Those messages are bit confusing as you won't really be switching to anything - you'll start a REPL of this type.

arichiardi: True, it was part of my first attempt

inf-clojure.el
arichiardi: I added it because I though you'd want to modify the `inf-clojure-program` only if it is set as a command , not if it is a cons cell, which indicates you don't want to launch anything...open to other ideas..

inf-clojure.el
arichiardi: Yeah, I'll change it

inf-clojure.el
arichiardi: And this is were I tell you that I am not an expert at all here, so please guide me on the best approach. For example how to set those `defcustom`s then? Should those be `buffer-local` instead?

inf-clojure.el
arichiardi: So I am beginning to understand a bit more, a `defcustom` is global and therefore this will be set once and forever right? So this becomes buffer-local? I need to read more on this...and thanks for reviewing!

inf-clojure.el
arichiardi: Don't know what it is, but I will check..

inf-clojure.el
bbatsov: Basically such variable such be declared as `defvar-local` instead of as `defcustoms`. There should probably be some defcustoms with the values for the different types of repls, but the actual thing that would be set in the code would be a local variable that's specific to the REPL. You can see in CIDER's code it makes heavy use of local variables.

bbatsov:
Btw, seems we can support planck properly using the same technique.

bbatsov:
In case you're wondering - #34 reminded me of this. Seems planck also needs different code for completion, ns switching and so on.

inf-clojure.el
arichiardi: Ok I was reading a bit more. By `defcustom` with the types you mean a `defcustom` that uses `:set` to set the local var? The only thing I am missing is how to set the buffer local from a `defcustom`. I am checking also `cider`.

inf-clojure.el
arichiardi: Fixed this in the other commit (which of course we can squash when things are good)

bbatsov: You should be setting and checking only buffer-locals. The defcustoms should simply exist as the values of the different flavours. This means you'll have one doc buffer-local and two defcustoms - one for clojure and one for lumo. Alternatively you can just have a buffer-local and hardcode the defcustom values, but this will make the code less flexible, so I wouldn't recommend it.

arichiardi: Oh I was afraid of this :smile: this means converting all the `defcustoms` I am using right?

arichiardi: On second thought, why are we exposing something like `inf-clojure-var-doc-command` anyways? It is kind of an implementation detail that we probably should hide it from user customization and therefore make it `defvar-local`.
As you case see above, a lot of those `defcustom`s need to be set to something else in case of `lumo` flavor, and will need to be changed again in case of `planck` flavor and "normal cljs" repl. So this would mean 1 `defvar-local` + 4 defcustom already for each var that needs customization...It might become unmanageable.

bbatsov: Whether those are defcustoms are hardcoded strings, the complexity of the code won't be changed that much. The advantage of using `defcustoms` is that users might be able to use some of the features on "unsupported" repls if they have the desire to customize the defcustoms. Or they can tweak the behaviour of something - e.g. use some different completion code. As I said - I can live with the code snippets for the various repls being hardcoded as well - it makes little difference to me.

arichiardi: OK then, I will begin the big refactoring soon-ish, just wanted to be sure it is what you were thinking 👍

inf-clojure.el
```diff
@@ -391,12 +384,20 @@ Fallback to `default-directory.' if not within a project."
...
+    (lumo (inf-clojure--lumo-repl-command))
+    (_ nil)))
+
 (defun inf-clojure-cmd (project-type)
   "Determine the command `inf-clojure' needs to invoke for the PROJECT-TYPE."
-  (pcase project-type
-    ("lein" inf-clojure-lein-cmd)
-    ("boot" inf-clojure-boot-cmd)
-    (_ inf-clojure-generic-cmd)))
+  (if-let ((flavor-command (inf-clojure-flavor)))
```
arichiardi: @bbatsov this is what I meant basically, if `inf-clojure-repl-flavor` is present, it takes precedence. It is either this is coming up with a project marker of our own, which I don't like that much.

inf-clojure.el
```diff
@@ -391,12 +384,20 @@ Fallback to `default-directory.' if not within a project."
           ((file-exists-p "build.boot") "boot")
           (t nil))))
 
+(defun inf-clojure-flavor ()
+  "Determine the flavor of the project."
+  (pcase inf-clojure-repl-flavor
+    (lumo (inf-clojure--lumo-repl-command))
+    (_ nil)))
```
arichiardi: Many cases here, if I had to sketch a Clojure data structure, I would do:

```
{:project-type     :boot, :lein, ...
 :project-flavor   :clj, :lumo, :planck, :cljs, ....}
```

bbatsov: The point is that you can't really detect this, you can only set it via some dir-local for instance. That's why I'd rather simply leave people to configure how to start the REPL, run some eval after the REPL starts to check what's its type and use the result to set its flavour. 

Alternatively there could be some commands like `inf-clojure-start-lumo` and so on, but I generally prefer to avoid having an excessive amount of commands.

So, I assume people would simply change `inf-clojure-lein-cmd` or whatever to a command starting a lumo repl if they want to do this.

arichiardi: Are you talking about the start command specifically or all the commands to customize?
Because it is a bit inconvenient imho to carry around a huge `.dir-locals.el` from project to project for instance, the local var seems like a good compromise, let me send an addition I made.

inf-clojure.el
arichiardi: I will remove this.

inf-clojure.el
```diff
@@ -566,6 +566,19 @@ The prefix argument SWITCH-TO-REPL controls whether to switch to REPL after the
   "Command to query inferior Clojure for a var's documentation."
   :type 'string)
 
+(defcustom inf-clojure-var-doc-command-lumo
+  "(lumo.repl/doc %s)"
+  "Lumo command to query inferior Clojure for a var's documentation."
+  :type 'string)
+
+(defun inf-clojure-var-doc-form ()
```
arichiardi: So I hope I got this right.

inf-clojure.el
```diff
@@ -261,12 +267,12 @@ whichever process buffer you want to use.")
...
 (put 'inf-clojure-mode 'mode-class 'special)
 
-(defcustom inf-clojure-repl-default-flavor 'clojure
+(defcustom inf-clojure-repl-default-flavor nil
   "Symbol to define your default REPL flavor.
 The default flavor is clojure, lumo is the other supported
 one."
   :type '(choice (const :tag "Lumo" lumo)
-                 (other :tag "Clojure" clojure))
+                 (other :tag "Project driven" nil))
```
arichiardi: This tag here clarifies that in case of missing flavor the project type will drive

inf-clojure.el
bbatsov: I'm fine with the naming. It's better indeed.

inf-clojure.el
```diff
@@ -599,6 +599,19 @@ If you are using flavors, it will pickup the most approapriate
   "Command to query inferior Clojure for completion candidates."
   :type 'string)
 
+(defcustom inf-clojure-completion-command-lumo
```
bbatsov: Probably we should use `form` in those names as well.

inf-clojure.el
```diff
@@ -778,7 +791,7 @@ See variable `inf-clojure-buffer'."
     (unwind-protect
         (let ((completion-snippet
                (format
-                inf-clojure-completion-command (substring-no-properties expr))))
+                inf-clojure-completion-form (substring-no-properties expr))))
```
bbatsov: This should have been a function invocation.

inf-clojure.el
```diff
@@ -250,6 +262,25 @@ one process, this does the right thing.  If you run multiple
 processes, you might need to change `inf-clojure-buffer' to
 whichever process buffer you want to use.")
 
+(defvar inf-clojure-mode-hook '()
```
bbatsov: Seems you didn't merge something cleanly. This definition and the subsequent `put` were deleted recently.

arichiardi: Yes it seems like I got this hunk while rebasing, thanks

arichiardi:
A question for @bbatsov and cc @anmonteiro. When I start `lumo` I see things like the following:

2017-03-05-194935_342x149_scrot


It looks like there is a mismatch in what `comint` can display and what lumo returns. If I enable the dumb terminal in `lumo` with `-d` is looks ok, but I don't have my commands echoed anymore.
I might need some hint on this.

arichiardi:
Ok probably I need to include the two characters in the filtered one on the output.

bbatsov:
I have no idea what those character are. I thought you tested this in the past and it was working ok then.

arichiardi:
Ok cleaned up the characters as well. At the moment it does the cleaning even when in other flavors, because I didn't have time to check how to set the buffer local var when launching the `*inf-clojure*` buffer. Anyways, I think that is a small fix for more expert eyes. It is in the last commit (dc63105
). I have learned more about Emacs today than in my whole life plus previous lives :smile:

inf-clojure.el
```diff
@@ -876,6 +943,46 @@ to suppress the usage of the target buffer discovery logic."
...
+          (comint-send-string proc (concat form "\n"))
+          ;; yey, loads of procedural code again
+          (while (and (not is-lumo)
+                      (not (string-match inf-clojure-prompt kept))
+                      (accept-process-output proc 2))
+            (setq is-lumo (string-match-p (concat form "\\Ca*true\\Ca*") kept))))
+      (set-process-filter proc prev-filter))
+    is-lumo))
+
+;; The defcustoms for the following are already ok:
```
bbatsov: Not sure if these comments belong here.

inf-clojure.el
```diff
@@ -876,6 +943,46 @@ to suppress the usage of the target buffer discovery logic."
...
+
+;; When we'll have tests, this is one
+;; (let ((form "(js/global.hasOwnProperty \"$$LUMO_GLOBALS\")")
+;;       (s "(js/global.hasOwnProperty \"$$LUMO_GLOBALS\")

\ntrue
\n[1G[0Jcljs.user=> [13G"))
+;;   (string-match (concat form "\\Ca*true\\Ca*") s))
 
+(defun inf-clojure--lumo-mode-p (proc)
```
bbatsov: I'd make this a generic function that takes as extra params some form and some expected result, so we could reuse it for other flavours.

inf-clojure.el
```diff
@@ -876,6 +943,46 @@ to suppress the usage of the target buffer discovery logic."
     (inf-clojure (inf-clojure-cmd (inf-clojure-project-type)))
     (rename-buffer target-buffer-name)))
 
-(provide 'inf-clojure)
+;;;; Lumo
+;;;; ====
+
+;; When we'll have tests, this is one
```
bbatsov: Guess you can remove this as the subsequent function actually tests for it.

inf-clojure.el
```diff
@@ -526,6 +567,19 @@ The prefix argument SWITCH-TO-REPL controls whether to switch to REPL after the
   "Command to query inferior Clojure for a var's documentation."
   :type 'string)
 
+(defcustom inf-clojure-var-doc-command-lumo
```
bbatsov: command -> form

inf-clojure.el
```diff
@@ -526,6 +567,19 @@ The prefix argument SWITCH-TO-REPL controls whether to switch to REPL after the
...
+  "(lumo.repl/doc %s)\n"
+  "Lumo command to query inferior Clojure for a var's documentation."
```
bbatsov: Same here.

inf-clojure.el
bbatsov: Add also change the name to `inf-clojure--lumo-p`.

inf-clojure.el
```diff
@@ -526,6 +567,19 @@ The prefix argument SWITCH-TO-REPL controls whether to switch to REPL after the
...
+  :type 'string)
+
+(defun inf-clojure-var-doc-form ()
+  "Return the form to query inferior Clojure for a var's documentation.
+If you are using flavors, it will pickup the most approapriate
+`inf-clojure-var-doc-command` variant."
```
bbatsov: command -> form

inf-clojure.el
```diff
@@ -176,13 +178,37 @@ often connecting to a remote REPL process."
...
 
-Alternative you can specify a TCP connection cons pair, instead
+Alternatively you can specify a TCP connection cons pair, instead
 of command, consisting of a host and port
 number (e.g. (\"localhost\" . 5555)).  That's useful if you're
 often connecting to a remote REPL process."
   :type '(choice (string)
                  (cons string integer)))
 
+(defvar-local inf-clojure-repl-flavor nil
```
bbatsov: Btw, probably we should simply use `type` instead of `flavor`. Seems more obvious to me.

inf-clojure.el
```diff
@@ -176,13 +178,37 @@ often connecting to a remote REPL process."
...
+
+(defun inf-clojure--probe-flavor (proc)
+  "Identifies the current REPL flavor for PROC."
+  (cond
+   ((inf-clojure--lumo-mode-p proc) 'lumo)
+   (t 'clojure)))
+
+(defun inf-clojure--set-flavor (proc)
+  "Set the flavor if has not already been set."
+  (when (not inf-clojure-repl-flavor)
```
bbatsov: when + not = unless :-)

inf-clojure.el
```diff
@@ -176,13 +178,37 @@ often connecting to a remote REPL process."
...
 often connecting to a remote REPL process."
   :type '(choice (string)
                  (cons string integer)))
 
+(defvar-local inf-clojure-repl-flavor nil
+  "Symbol to define your REPL flavor.
+Its root binding is nil and it can be further customized using
+either `setq-local` or an entry in `.dir-locals.el`." )
+
+(defun inf-clojure--probe-flavor (proc)
```
bbatsov: detect type perhaps?

bbatsov:
The PR is in pretty good shape already. I've added a few small remarks and that's pretty much all.

You'll also have to update the changelog and the README - maybe you can add a small section on lumo there.

inf-clojure.el
arichiardi: Right! Emacs has that 😀

inf-clojure.el
arichiardi: Yeah probably they are issues to open

inf-clojure.el
arichiardi: It is maybe better, I so like flavor I will need to find an open source project where to use it 😀 Only Gradle was so bold as to actually add it as concept 😁

arichiardi:
I renamed all of the vars, is it ok or you prefer another PR for that?

arichiardi:
Ok I made the require changes, REPL `type` is in. The renaming of the commands to `-forms` is not in here anymore, I will open another PR once this is in.

bbatsov:
OK, guess you can update the changelog and the README in the second PR.

Btw, if you have any `planck` familiarity it'd be great to add support for it as well. Now this should be trivial.

arichiardi:
There is still a case when the repl is not detected, I have a patch already but I want to try it thoroughly

PR #46, opened 2017-03-03
Add defcustom `inf-clojure-prompt-when-set-ns`

arichiardi:
The patch allow the user to disable the request "Set ns: ..." on C-c
M-n. The default is to always prompt.

(No need?) -> You've updated the readme (if adding/changing user-visible functionality)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### New Features
 
 * [#34](https://github.com/clojure-emacs/inf-clojure/pull/34): Add support for socket REPL connections.
+* [#46](https://github.com/clojure-emacs/inf-clojure/pull/46): Make it possible to disable prompt on `inf-clojure-set-ns`.
 * New interactive command `inf-clojure-display-version`.
 
 ### Bugs Fixed
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -179,6 +179,12 @@ This should usually be a combination of `inf-clojure-prompt' and
   :type 'regexp
   :group 'inf-clojure)
 
+(defcustom inf-clojure-prompt-on-set-ns t
+  "Controls whether to prompt when switching namespace."
+  :type '(choice (const :tag "always" t)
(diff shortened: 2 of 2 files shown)


inf-clojure.el
```diff
@@ -179,6 +179,12 @@ This should usually be a combination of `inf-clojure-prompt' and
   :type 'regexp
   :group 'inf-clojure)
 
+(defcustom inf-clojure-prompt-when-set-ns t
```
bbatsov: when -> on

inf-clojure.el
```diff
@@ -179,6 +179,12 @@ This should usually be a combination of `inf-clojure-prompt' and
...
+  "Controls whether to prompt by default when switching namespace."
```
bbatsov: "by default" seems redundant here

inf-clojure.el
```diff
@@ -620,8 +626,14 @@ See variable `inf-clojure-ns-vars-command'."
 
 (defun inf-clojure-set-ns (ns)
   "Set the ns of the inferior Clojure process to NS.
-Defaults to the ns of the current buffer."
-  (interactive (inf-clojure-symprompt "Set ns to" (clojure-find-ns)))
+Defaults to the ns of the current buffer, always prompting before
+setting, unless `inf-clojure-prompt-when-set-ns` is \"never\" or
```
bbatsov: I'm pretty sure "never" won't work - this is just a tag in the customize UI.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### New Features
 
 * [#34](https://github.com/clojure-emacs/inf-clojure/pull/34): Add support for socket REPL connections.
+* [#46](https://github.com/clojure-emacs/inf-clojure/pull/46): Optinally disable prompt on `inf-clojure-set-ns`.
```
bbatsov: Make it possible to disable prompt...

PR #51, opened 2017-03-08
Add inf-clojure-prompt-on-commands and closes #48

### LINKED ISSUE #48: Avoid prompting on doc, source, argument inspection
Hi Bozhidar,

I would like to add an option that disables prompting at every `inf-clojure-show-*` function.
Cider's workflow is way more fluid. I was wondering whether we should have a `defcustom` for each and every command or a global one. Probably the latter.

Thanks a lot!

arichiardi:
The patch introduces a way to globally disable prompting. It deletes
`inf-clojure-prompt-when-set-ns`.

- [X] You've updated the README

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -238,12 +238,6 @@ This should usually be a combination of `inf-clojure-prompt' and
 `inf-clojure-subprompt'."
   :type 'regexp)
 
-(defcustom inf-clojure-prompt-on-set-ns t
-  "Controls whether to prompt when switching namespace."
-  :type '(choice (const :tag "always" t)
-                 (const :tag "never" nil))
-  :package-version '(inf-clojure . "2.0.0"))
-
 (defcustom inf-clojure-repl-use-same-window nil
   "Controls whether to display the REPL buffer in the current window or not."
   :type '(choice (const :tag "same" t)
@@ -722,22 +716,29 @@ The value is nil if it can't find one."
 ;;; Documentation functions: var doc and arglist.
 ;;; ======================================================================
 
-(defun inf-clojure-show-var-documentation (var)
+(defun inf-clojure-show-var-documentation (prompt-for-symbol)
   "Send a form to the inferior Clojure to give documentation for VAR.
-See function `inf-clojure-var-doc-form'."
-  (interactive (inf-clojure-symprompt "Var doc" (inf-clojure-var-at-pt)))
-  (comint-proc-query (inf-clojure-proc) (format (inf-clojure-var-doc-form) var)))
+See function `inf-clojure-var-doc-form'.  When invoked with a
+prefix argument PROMPT-FOR-SYMBOL, it prompts for a symbol name."
+  (interactive "P")
+  (let ((var (if prompt-for-symbol
+                 (car (inf-clojure-symprompt "Var doc" (inf-clojure-var-at-pt)))
+               (inf-clojure-var-at-pt))))
+    (comint-proc-query (inf-clojure-proc) (format (inf-clojure-var-doc-form) var))))
 
-(defun inf-clojure-show-var-source (var)
-  "Send a form to the inferior Clojure to give source for VAR.
-See variable `inf-clojure-var-source-form'."
-  (interactive (inf-clojure-symprompt "Var source" (inf-clojure-var-at-pt)))
-  (comint-proc-query (inf-clojure-proc) (format inf-clojure-var-source-form var)))
+(defun inf-clojure-show-var-source (prompt-for-symbol)
+  "Send a command to the inferior Clojure to give source for VAR.
+See variable `inf-clojure-var-source-form'.  When invoked with a
+prefix argument PROMPT-FOR-SYMBOL, it prompts for a symbol name."
+  (interactive "P")
+  (let ((var (if prompt-for-symbol
+                 (car (inf-clojure-symprompt "Var source" (inf-clojure-var-at-pt)))
+               (inf-clojure-var-at-pt))))
+    (comint-proc-query (inf-clojure-proc) (format inf-clojure-var-source-form var))))
 
 (defun inf-clojure-arglist (fn)
   "Send a query to the inferior Clojure for the arglist for function FN.
 See variable `inf-clojure-arglist-form'."
-  (interactive (inf-clojure-symprompt "Arglist" (inf-clojure-fn-called-at-pt)))
   (let* ((proc (inf-clojure-proc))
          (comint-filt (process-filter proc))
          (kept "")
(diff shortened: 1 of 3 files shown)


arichiardi:
Actually this does not make proper sense in terms of usability, `apropos` for instance has to prompt I guess. Maybe it is better if we come up with a list of commands that need to prompt for input.

arichiardi:
On second though I would probably always prompt for `apropos` and that is it...even better would some to invert the behavior of `inf-clojure-prompt-on-commands` with a `C-u` prefix

bbatsov:
> Actually this does not make proper sense in terms of usability, apropos for instance has to prompt I guess. Maybe it is better if we come up with a list of commands that need to prompt for input.

Yeah, appropos shouldn't operate at all on this symbol at point. Guess we should emulate the cider behaviour for it.

> On second though I would probably always prompt for apropos and that is it...even better would some to invert the behavior of inf-clojure-prompt-on-commands with a C-u prefix

Yeah, it'd be best to add prefix arg support for all commands can prompt or not.

bbatsov:
You'll have to rebase this, btw.

arichiardi:
Don't know how you understood the latter msg, it looks like my fingers did not want to write some words 😀

arichiardi:
Is `inf-clojure-arglist` supposed to prompt? It seems to me that it should not, as it is only called by eldoc and it is guarded by the `inf-clojure-show-arglist` command. Just wanted a confirmation.

bbatsov:
> Is inf-clojure-arglist supposed to prompt? It seems to me that it should not, as it is only called by eldoc and it is guarded by the inf-clojure-show-arglist command. Just wanted to confirm.

This command can probably be removed. It was added before the mode had `eldoc` support as a simple alternative to it. Seems pretty useless to me right now.

arichiardi:
Ok I will remove it

inf-clojure.el
```diff
@@ -725,19 +731,24 @@ The value is nil if it can't find one."
 (defun inf-clojure-show-var-documentation (var)
   "Send a form to the inferior Clojure to give documentation for VAR.
 See function `inf-clojure-var-doc-form'."
-  (interactive (inf-clojure-symprompt "Var doc" (inf-clojure-var-at-pt)))
-  (comint-proc-query (inf-clojure-proc) (format (inf-clojure-var-doc-form) var)))
+  (interactive "P")
+  (let ((var (if (inf-clojure--should-prompt-on-commands var)
```
bbatsov: Seems to me you're not using your helper function correctly. I assume it was intended to be used with some prefix argument.

inf-clojure.el
```diff
@@ -238,11 +238,17 @@ This should usually be a combination of `inf-clojure-prompt' and
...
-(defcustom inf-clojure-prompt-on-set-ns t
+(defcustom inf-clojure-prompt-on-commands t
   "Controls whether to prompt when switching namespace."
   :type '(choice (const :tag "always" t)
                  (const :tag "never" nil))
-  :package-version '(inf-clojure . "2.0.0"))
+  :package-version '(inf-clojure . "2.0.0")
+  :safe 'booleanp)
+
+(defun inf-clojure--should-prompt-on-commands (&optional invert)
```
bbatsov: prompt-on-commands-p

inf-clojure.el
```diff
@@ -238,11 +238,17 @@ This should usually be a combination of `inf-clojure-prompt' and
 `inf-clojure-subprompt'."
   :type 'regexp)
 
-(defcustom inf-clojure-prompt-on-set-ns t
+(defcustom inf-clojure-prompt-on-commands t
```
bbatsov: That's a bit vague. Maybe "prompt-for-symbol-at-point" or something. And you forgot to update the docstring.

inf-clojure.el
arichiardi: Can you expand on this? It is working fine here, prompting when the function returns true.

arichiardi: still valid, sorry for the mess :wink:

inf-clojure.el
arichiardi: the new name would require to flip the logic and rename also the other functions...also maybe in the future not all the commands will take symbol at point so I was wondering if we could leave it like this.

bbatsov: "like this" meaning what?

arichiardi: Meaning `inf-clojure-prompt-on-commands`

bbatsov: I can live with this.

arichiardi: Ok let me know if this requires more work

inf-clojure.el
```diff
@@ -725,19 +731,24 @@ The value is nil if it can't find one."
...
 
 (defun inf-clojure-show-var-source (var)
-  "Send a form to the inferior Clojure to give source for VAR.
-See variable `inf-clojure-var-source-form'."
-  (interactive (inf-clojure-symprompt "Var source" (inf-clojure-var-at-pt)))
-  (comint-proc-query (inf-clojure-proc) (format inf-clojure-var-source-form var)))
+  "Send a command to the inferior Clojure to give source for VAR.
+See variable `inf-clojure-var-source-command'."
+  (interactive "P")
+  (let ((var (if (inf-clojure--prompt-on-commands-p var)
```
bbatsov: You should add the prefix argument to the function signature and pass it here, instead of `var`.

arichiardi: Isn't that just a name (symbol)? Btw there is the same thing in `cider`. I mean, it is working this way, maybe there is a better way I am not getting...Can you show me a simple piece of code?

bbatsov: cider uses the magic `current-prefix-arg` variable for this. 

```
(defun cider--find-dwim-interactive (prompt)
  "Get interactive arguments for jump-to functions using PROMPT as needed."
  (if (cider--prompt-for-symbol-p current-prefix-arg)
      (list
       (cider-read-from-minibuffer prompt (thing-at-point 'filename)))
    (list (or (thing-at-point 'filename) ""))))  ; No prompt.

(defun cider--invert-prefix-arg (arg)
  "Invert the effect of prefix value ARG on `cider-prompt-for-symbol'.

This function preserves the `other-window' meaning of ARG."
  (let ((narg (prefix-numeric-value arg)))
    (pcase narg
      (16 -1)   ; empty empty -> -
      (-1 16)   ; - -> empty empty
      (4 nil)   ; empty -> no-prefix
      (_ 4)))) ; no-prefix -> empty

(defun cider--prefix-invert-prompt-p (arg)
  "Test prefix value ARG for its effect on `cider-prompt-for-symbol`."
  (let ((narg (prefix-numeric-value arg)))
    (pcase narg
      (16 t) ; empty empty
      (4 t)  ; empty
      (_ nil))))

(defun cider--prompt-for-symbol-p (&optional prefix)
  "Check if cider should prompt for symbol.

Tests againsts PREFIX and the value of `cider-prompt-for-symbol'.
Invert meaning of `cider-prompt-for-symbol' if PREFIX indicates it should be."
  (if (cider--prefix-invert-prompt-p prefix)
      (not cider-prompt-for-symbol) cider-prompt-for-symbol))
```

bbatsov: Btw, for consistency you can use the same defcustom name - `inf-clojure-prompt-for-symbol`.

bbatsov: One more thing - by moving out the var init from `interactive` you're going to break those functions when invoked without parameters. I'm guessing you didn't really test this code. :-) You'll have to make params like `var` optional and prompt for them only when they are not passed explicitly (or remove them completely).

arichiardi: I will paste a gif showing it OK? If I hadn't tested it I would not even have opened a PR 😀 maybe I did not test some case, but all the interactive calls are working fine this way. Do you mean a call with no params from `ielm` or call with no params on `M-x`. The latter works but I will confirm it again today.

arichiardi: Also, here is the excerpt I copied from cider: https://github.com/clojure-emacs/cider/blob/1ce1d9998fc756d3dbbb491cc6f4f0c4c397b450/cider-doc.el#L208

bbatsov: Read carefully the code you copied - `arg` there is the prefix arg, not the the symbol at point (like in your case). I guess what you did works because when you removed the `var` related code from `interactive` there are two bindings for `var` within the code, but this is definitely wrong. If you want to do it this way you should rename var to `prompt-for-symbl` or something.

arichiardi: Maybe I am used to Clojure were you cannot have two bindings at the same time, the first var symbol is bound to the arg, the second is in the let scope and is bound to the result of the function call. In Clojure you take advantage of this to override and replace the value in input with the one you calculate and want to use in the scope of the let.

I don't know emacs' guts enough to know it's different semantics about var binding and it might be confusing so I think I am going to rename it.

arichiardi: Sorry to bother on this again, I am reading [this](https://www.gnu.org/software/emacs/manual/html_node/elisp/Local-Variables.html) and I really don't see any difference with Clojure and no reason why the above is wrong:

> A variable can have more than one local binding at a time (e.g., if there are nested let forms that bind the variable). The current binding is the local binding that is actually in effect. It determines the value returned by evaluating the variable symbol, and it is the binding acted on by setq.

Again, you ae the boss so I will rename it 😀

inf-clojure.el
```diff
@@ -725,19 +731,24 @@ The value is nil if it can't find one."
 (defun inf-clojure-show-var-documentation (var)
   "Send a form to the inferior Clojure to give documentation for VAR.
 See function `inf-clojure-var-doc-form'."
-  (interactive (inf-clojure-symprompt "Var doc" (inf-clojure-var-at-pt)))
-  (comint-proc-query (inf-clojure-proc) (format (inf-clojure-var-doc-form) var)))
+  (interactive "P")
+  (let ((var (if (inf-clojure--prompt-on-commands-p var)
```
bbatsov: You're killing me, buddy. :-) `var` is a prefix arg at this point, but after the `if` returns it's overridden - it becomes the symbol at point. I assumed this was apparent. I find this extremely confusing, so I'd change the parameter name to something else, the `let`-binding is fine. Hopefully, I managed to make myself clear this time around.

arichiardi: Sorry bro, I am learning right? 😀 Yeah no it was clear, I am going to change that and also use the more robust invert you posted from cider above.

bbatsov: 👍

arichiardi:
Ok done, waiting for you further remarks.

bbatsov:
Btw, I meant you should change the parameter names to something `invert-prompting` or whatever, there was no need to change the let-binding names. You should also mention in each interactive command what does this prefix argument to do.

inf-clojure.el
```diff
@@ -238,11 +238,23 @@ This should usually be a combination of `inf-clojure-prompt' and
...
-(defcustom inf-clojure-prompt-on-set-ns t
+(defcustom inf-clojure-prompt-on-commands t
   "Controls whether to prompt when switching namespace."
   :type '(choice (const :tag "always" t)
                  (const :tag "never" nil))
-  :package-version '(inf-clojure . "2.0.0"))
+  :package-version '(inf-clojure . "2.0.0")
+  :safe 'booleanp)
+
+(defun inf-clojure--prompt-on-commands-p (&optional arg)
```
bbatsov: Isn't copying this an overkill? It was needed in CIDER, because some commands had prefix arguments of their own, but here it seems pointless to make it this complex.

Frankly I wonder if we even need this `defcustom` we can make all the commands not prompt by default and to prompt with a prefix argument (or when there's nothing at point). It would be simple and would get the job done.

arichiardi: Agree on this! I will revert back to the previous version and abolish the defcustom.

inf-clojure.el
```diff
@@ -722,22 +716,29 @@ The value is nil if it can't find one."
...
+PROMPT-FOR-SYMBOL it prompts for a symbol name."
+  (interactive "P")
+  (let ((var (if prompt-for-symbol
+                 (car (inf-clojure-symprompt "Var doc" (inf-clojure-var-at-pt)))
+               (inf-clojure-var-at-pt))))
+    (comint-proc-query (inf-clojure-proc) (format (inf-clojure-var-doc-form) var))))
+
+(defun inf-clojure-show-var-source (prompt-for-symbol)
+  "Send a command to the inferior Clojure to give source for VAR.
+See variable `inf-clojure-var-source-form', with non-nil
```
bbatsov: I'd use a separate sentence for the prefix arg everywhere.

When invoked with a prefix argument `prompt-for-symbol`....

PR #52, opened 2017-03-08
Fix a problem with pattern matching on inf-clojure-repl-type

arichiardi:
It was never matching on the "default" branch with symbols, we switch to
strings.

---

I break it, I fix it :smile:

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -577,7 +577,7 @@ The prefix argument SWITCH-TO-REPL controls whether to switch to REPL after the 
 If you are using REPL types, it will pickup the most approapriate
 `inf-clojure-var-doc-form` variant."
   (pcase inf-clojure-repl-type
-    (lumo inf-clojure-var-doc-form-lumo)
+    (`lumo inf-clojure-var-doc-form-lumo)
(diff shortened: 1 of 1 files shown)


inf-clojure.el
```diff
@@ -577,7 +577,7 @@ The prefix argument SWITCH-TO-REPL controls whether to switch to REPL after the
 If you are using REPL types, it will pickup the most approapriate
 `inf-clojure-var-doc-form` variant."
   (pcase inf-clojure-repl-type
-    (lumo inf-clojure-var-doc-form-lumo)
```
bbatsov: You simply meant write ``lumo` here.

arichiardi: I tried that backtick in `ielm` and it did not work, it was always going to the else branch. Also in `cider` I have never seen using symbols so I settled for strings.

bbatsov: I'm 100% that this is the correct syntax for `pcase` symbols and it's generally better to use symbols over strings (strings are not interned in Elisp). If cider uses only strings in such cases it's just accidental.

arichiardi: So one backtick in front ? I will try again

PR #55, opened 2017-03-12
Kill inf-clojure-var-at-point

arichiardi:
Now every function calls the more reliable inf-clojure-symbol-at-point.

Please let me know if there was some valid reason not to do so, but it was not behaving really well, especially in conjunction with all these filters we have in place.

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -698,17 +698,6 @@ The value is nil if it can't find one."
             (and (symbolp obj) obj))))
     (error nil)))
 
-
-;;; Adapted from variable-at-point in help.el.
-(defun inf-clojure-var-at-pt ()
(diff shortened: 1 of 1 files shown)


arichiardi:
Please hold on this one. I have noticed that the function is probably patching some weird behaviour introduced by filtering the output. I need to investigate more.

bbatsov:
> Please hold on this one. I have noticed that the function is probably patching some weird behaviour introduced by filtering the output. I need to investigate more.

I don't see anything related to comint in this function. Likely I just copy/pasted it from `inferior-lisp.el`.

arichiardi:
Yes I tried this on both `clojure` and `lumo` repl, it does not seem to have regressions.

PR #59, opened 2017-03-14
Fix repl type sync

arichiardi:
I opened this PR basically to raise a flag around synchronicity and asking for what is the best way to solve them. It is a very sporadic problem I get when I try to detect the REPL type while waiting for a REPL detection that finishes. As you can imagine these things are difficult to reproduce and I'd like to ask folks to try `inf-clojure` without this patch in `lumo` and report what happens.

I found http://blog.jorgenschaefer.de/2014/05/race-conditions-in-emacs-process-filter.html and put a lock in there, which is a hack.

Thanks!

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -195,17 +195,23 @@ often connecting to a remote REPL process."
 Its root binding is nil and it can be further customized using
 either `setq-local` or an entry in `.dir-locals.el`." )
 
-(defun inf-clojure--detect-type (proc)
+(defvar inf-clojure--repl-type-lock nil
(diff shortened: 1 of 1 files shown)


arichiardi:
The PR includes the revert only for convenience. Once you revert on master I'll rebase.

arichiardi:
I have to say that without this is everything looks broken for `lumo` :smile:

bbatsov:
You'll have to rebase.

arichiardi:
Ok restricted the scope of this PR

PR #62, opened 2017-03-18
Add dir support in Lumo

arichiardi:
This patch adds `inf-clojure-ns-vars-form-lumo` and a function that
 return the correct `inf-clojure-ns-vars-form` variant given the REPL
 type.

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -133,7 +133,7 @@ mode.  Default is whitespace followed by 0 or 1 single-letter colon-keyword
         ["Show arglist" inf-clojure-show-arglist t]
         ["Show documentation for var" inf-clojure-show-var-documentation t]
         ["Show source for var" inf-clojure-show-var-source t]
-        ["Show vars in ns" inf-clojure-show-ns-varst]
+        ["Show vars in ns" inf-clojure-show-ns-vars t]
         ["Apropos" inf-clojure-apropos t]
         ["Macroexpand" inf-clojure-macroexpand t]
         "--"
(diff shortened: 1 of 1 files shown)


inf-clojure.el
```diff
@@ -773,7 +787,7 @@ prefix argument PROMPT-FOR-NS, it prompts for a namespace name."
   (let ((ns (if prompt-for-ns
                 (car (inf-clojure-symprompt "Ns vars" (clojure-find-ns)))
               (clojure-find-ns))))
-    (comint-proc-query (inf-clojure-proc) (format inf-clojure-ns-vars-form ns))))
+    (inf-clojure--send-string (inf-clojure-proc) (format (inf-clojure-ns-vars-form) ns))))
```
bbatsov: Is this replacement equivalent? I'm not 100% sure about this. I also don't think it's needed.

arichiardi:
It is there for consistency and also because the inf-clojure--send-string  actually sets up the repl type.


-- 

inf-clojure.el
arichiardi: Sorry the answer from the email got pasted below.

arichiardi: Let me try again with the old one and see

bbatsov: I just wonder what's the difference between `send-string` and `proc-query`. I said I don't think it's needed as likely setting the ns is not going to be first thing you do with the repl, but I might be mistaken in my assumption.

arichiardi: This is a very good point and the subject of #55 , who knows what is the user going to do? Now we set the repl type in the right places, even in more than necessary maybe. Better be safe 😀

However it looks like there is a reason why it is like this: https://github.com/emacs-mirror/emacs/blob/master/lisp/comint.el#L2899: to receive something back from the function instead of showing it in the buffer.

I have also discovered that when you set a process filter (which we do there) the output is not pasted to the buffer.

So please let me to some further trial on this one.

arichiardi:
Ok reverted to old behavior. This patch now is ready, not that the `dir` `lumo` feature has not been released yet.

PR #64, opened 2017-03-21
Fix newline append on eval functions

arichiardi:
A fix was necessary for a regression caused by the fact that now
`inf-clojure--send-string` adds a newline at the end of the string it
sends to the process.

-----------------


arichiardi:
Also wanted to ask an opinion: currectly `inf-clojure-load-form` it is not duplicated for `lumo`. I was thinking that forms that are already good in both repl types should not be duplicated.
I am having second thoughts though, because a) it is not consistent b) it is not configurable anymore.
So what do you think? Shall duplicate all the forms? I'd say yes.

bbatsov:
> So what do you think? Shall duplicate all the forms? I'd say yes.

Yes.

PR #66, opened 2017-03-24
Fix #56 Add special handling for Planck REPLS

### LINKED ISSUE #56: Add special handling for planck REPLs
Basically we need to extend what was done in #44 to cover planck as well, so all the `inf-clojure` functionality would work with planck properly.

This should be pretty easy and I would have done it myself if I knew the first thing about planck. :-)

slipset:
Added support for the Planck REPL

Thanks!

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -237,7 +237,7 @@ Clojure to load that file."
 (define-obsolete-variable-alias 'inf-clojure-load-command 'inf-clojure-load-form "2.0.0")
 
 (defcustom inf-clojure-load-form-lumo "(clojure.core/load-file \"%s\")"
-  "Planck format-string for building a Clojure expression to load a file.
+  "Lumo format-string for building a Clojure expression to load a file.
 This format string should use `%s' to substitute a file name and
 should result in a Clojure form that will be sent to the inferior
 Clojure to load that file."
@@ -282,7 +282,7 @@ This should usually be a combination of `inf-clojure-prompt' and
   :package-version '(inf-clojure . "2.0.0"))
 
 (defvar inf-clojure-buffer nil
-  "The current ‘inf-clojure’ process buffer.
+  "The current inf-clojure process buffer.
 
 MULTIPLE PROCESS SUPPORT
 ===========================================================================
@@ -681,7 +681,7 @@ If you are using REPL types, it will pickup the most approapriate
   :package-version '(inf-clojure . "2.0.0"))
 
 (defcustom inf-clojure-completion-form-planck
-  "(doall (map str (planck.repl/get-completions \"%s\")))\n"
+  "(planck.repl/get-completions \"%s\")\n"
   "Planck form to query inferior Clojure for completion candidates."
   :type 'string
   :package-version '(inf-clojure . "2.0.0"))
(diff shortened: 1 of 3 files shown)


inf-clojure.el
```diff
@@ -236,7 +237,15 @@ Clojure to load that file."
 (define-obsolete-variable-alias 'inf-clojure-load-command 'inf-clojure-load-form "2.0.0")
 
 (defcustom inf-clojure-load-form-lumo "(clojure.core/load-file \"%s\")"
-  "Format-string for building a Clojure expression to load a file.
+  "Planck format-string for building a Clojure expression to load a file.
```
arichiardi: Maybe little typo here, shouldn't it be for Lumo? I have probably forgot to update this one myself :smile:

inf-clojure.el
```diff
@@ -272,7 +282,7 @@ This should usually be a combination of `inf-clojure-prompt' and
   :package-version '(inf-clojure . "2.0.0"))
 
 (defvar inf-clojure-buffer nil
-  "The current inf-clojure process buffer.
+  "The current ‘inf-clojure’ process buffer.
```
arichiardi: Oh just out of curiosity, why those forward ticks?

inf-clojure.el
```diff
@@ -649,12 +680,19 @@ If you are using REPL types, it will pickup the most approapriate
   :type 'string
   :package-version '(inf-clojure . "2.0.0"))
 
+(defcustom inf-clojure-completion-form-planck
+  "(doall (map str (planck.repl/get-completions \"%s\")))\n"
```
arichiardi: this `(doall (map str ...` was necessary because lumo returns `#js`. Does planck return it as well?

inf-clojure.el
```diff
@@ -685,6 +730,20 @@ If you are using REPL types, it will pickup the most approapriate
   :type 'string
   :package-version '(inf-clojure . "2.0.0"))
 
+(defcustom inf-clojure-set-ns-form-planck
+  "(in-ns '%s)\n"
```
arichiardi: no namespace here?

slipset: Seems like `in-ns` lives in top-level
```Clojure
cljs.user=> (doc in-ns)
-------------------------
in-ns
([name])
REPL Special Function
  Sets *cljs-ns* to the namespace named by the symbol, creating it if needed.
nil
cljs.user=> (doc apropos)
-------------------------
planck.repl/apropos
([str-or-pattern])
Macro
  Given a regular expression or stringable thing, return a seq of all
  public definitions in all currently-loaded namespaces that match the
  str-or-pattern.
```

inf-clojure.el
slipset: 8929305

inf-clojure.el
slipset: checkdoc suggested it. removed
8929305

inf-clojure.el
slipset: 8929305

slipset:
Thanks for the review @arichiardi!

bbatsov:
👍

inf-clojure.el
mfikes: Right, Planck's `in-ns` is a REPL special (just like in the ClojureScript REPL). It has not namespace.

PR #69, opened 2017-04-12
[Fix #63] Avoid spurious output by using buffer redirection

### LINKED ISSUE #63: Spurious output when first opened in lumo
## Expected behavior

When first opening the repl and cursor is on code buffer, only `cljs.user =>` should appear.

## Actual behavior

The output of a command (for me it is arglists because the cursor was after fn symbol, but this varies) is printed in the `*inf-clojure*` buffer.

`C-c C-z` with popup window enabled and move the cursor around in the code buffer.

or in `ielm`:

```
ELISP> (setq inf-clojure-repl-type nil)
nil
ELISP> (inf-clojure-arglist 'defn)
nil
```

*Latest master*

### Lein/Boot version

*Lumo latest master*

Reminder to self :sweat: :grimacing: :grinning:

arichiardi:
I will need to fix the arglists function as well, it does not work properly for edge cases, I already have some code ready, but I figured I should do another PR.

Also, the same approach should be used for completions, but completions require more work that is why I avoided putting it here.

-----------------

Thanks!

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -215,6 +215,12 @@ It requires a REPL PROC for inspecting the correct type."
       (setq inf-clojure-repl-type (inf-clojure--detect-repl-type proc))
     inf-clojure-repl-type))
 
+(defun inf-clojure--sanitize-command (command)
+  "Sanitize COMMAND for sending it to a process.
+An example of things that this function does is to add a final
+newline at the end of the form."
+  (concat (string-trim-right command) "\n"))
+
 (defun inf-clojure--send-string (proc string)
   "A custom `comint-input-sender` / `comint-send-string`.
 It performs the required side effects on every send for PROC and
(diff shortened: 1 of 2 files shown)


arichiardi:
The problem with completions is that you have to post-process them in different ways depending on the repl type. For instance lumo returns json, which emacs can read. So completions need a better handling.

inf-clojure.el
```diff
@@ -884,22 +884,46 @@ safe to call only from inside `inf-clojure-arglist`."
...
-    (unwind-protect
-        (let ((eldoc-snippet (format (inf-clojure-arglists-form) fn)))
-          (inf-clojure--send-string proc eldoc-snippet)
-          (while (and (not (string-match inf-clojure-prompt kept))
-                      (accept-process-output proc 2)))
-          (setq eldoc (inf-clojure-match-arglists eldoc-snippet kept)))
-      (set-process-filter proc comint-filt))
-    eldoc))
+  (let ((eldoc-snippet (format (inf-clojure-arglists-form) fn)))
+    (inf-clojure-results-from-process (inf-clojure-proc) (concat eldoc-snippet "\n"))))
```
bbatsov: Why don't you ensure the command ends with newline in `inf-clojure-results-from-process`?

inf-clojure.el
```diff
@@ -1145,20 +1169,9 @@ to suppress the usage of the target buffer discovery logic."
...
-        ;; use the real comind-send-string in order to avoid stack overflows
-        (comint-send-string proc (concat form "\n"))
-      ;; yey, loads of procedural code again
-      (while (and (not is-matching)
-                  (not (string-match inf-clojure-prompt kept))
-                  (accept-process-output proc 2))
-        (setq is-matching (funcall match-p kept)))
-      (set-process-filter proc prev-filter))
-    is-matching))
+  (thread-last
```
bbatsov: The use of `thread-last` here seems pointless to me.

inf-clojure.el
arichiardi: Yeah maybe that's better

inf-clojure.el
arichiardi: The command was so long that is why I used it, I can remove it

arichiardi:
Fixed the issue, I will need another sweep for fixing the problems with arglists and completions once and for all. Also, we could remove all the `\n` at the end of the form now, but I haven't included it here.

PR #81, opened 2017-06-02
add callback argument to lumo get-completions

hlolli:
lumo.repl/get-completions takes two arguments.
https://github.com/anmonteiro/lumo/blob/master/src/cljs/snapshot/lumo/repl.cljs#L1252
So I added map str function into the callback, now I'm getting completions suggestions from company-mode.

-----------------

Thanks!

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -726,7 +726,10 @@ If you are using REPL types, it will pickup the most approapriate
 (define-obsolete-variable-alias 'inf-clojure-completion-command 'inf-clojure-completion-form "2.0.0")
 
 (defcustom inf-clojure-completion-form-lumo
-  "(doall (map str (lumo.repl/get-completions \"%s\")))"
+  "(let [ret (atom)]
+     (lumo.repl/get-completions \"%s\"
+       (fn [res] (reset! ret (map str res))))
+     @ret)"
   "Lumo form to query inferior Clojure for completion candidates."
   :type 'string
   :package-version '(inf-clojure . "2.0.0"))


hlolli:
Yes, the completions work, but arglists aren't appearing while scrolling the completion list via company. Could you test this and the other pull request I made, or look over it. I see that on lumo side, sometimes lumo gives wrong completions, there is to say, all possible completions, in my case after ) character.

arichiardi:
So the callback thing is a pain (notice the `nil` at the end because of the print) :

```
cljs.user=> (lumo.repl/get-completions "redu" #(print (map str %)))
(reduce reduce-kv reduceable? reduced reduced? reductions)nil
```

hlolli:
That nil should go trough the elisp `read` function and become ignored when it hit the case statement. I regret not making two pull requests as I see this are two different problems interleaved with one another.

arichiardi:
I'd solve it this way:

```
(let [hack (atom)]
  (lumo.repl/get-completions "redu" #(swap! hack (fn [r] (map str %))))
  @hack)
```

hlolli:
Very good! Agree, I update the pull request and use your snippet :smile:

arichiardi:
Fixed.

hlolli:
Are you commenting on the correct pull requests? This fixes it or you've fixed it?

arichiardi:
No sorry I have fixed the comment with the snippet, I cannot fix the PR :smile:

arichiardi:
This has another problem,I will raise it in `lumo`.

hlolli:
PR updated.

arichiardi:
As per the linked `lumo` issue, this avoid pretty pretting:

```
cljs.user=> (do (set! lumo.repl/*pprint-results* false) (let [hack (atom)] (lumo.repl/get-completions "redu" #(reset! hack (map str %))) @hack))
("Reduced" "reduce" "reduce-kv" "reduceable?" "reduced" "reduced?" "reductions")
```

inf-clojure.el
```diff
@@ -726,7 +726,10 @@ If you are using REPL types, it will pickup the most approapriate
 (define-obsolete-variable-alias 'inf-clojure-completion-command 'inf-clojure-completion-form "2.0.0")
 
 (defcustom inf-clojure-completion-form-lumo
-  "(doall (map str (lumo.repl/get-completions \"%s\")))"
+  "(let [ret (atom)]
```
arichiardi: If you set it to:

`(do (set! lumo.repl/*pprint-results* false) (let [hack (atom)] (lumo.repl/get-completions "redu" #(reset! hack (map str %))) @hack))`

and squash all commits into one, we are good :smile:

hlolli: I'll do that and squash. But isn't this a bit mean to someone who wants to prettyprint the results when emacs is always setting it to false?

arichiardi:
Uhm, Antonio suggested it but you are right, it is global...Maybe this is the moment when we introduce a post processing step to completion. First I would try if it works with multiline lists, if not we have a fn that makes everything single line but we need to introduce some glue for post processing..

hlolli:
There are functions that can do that, but as far as I see, it returns a list of strings which is read as emacs data, so newline should be irrelevant, in company code completions list it makes no difference at least.

hlolli:
Ok I changed `swap!` to `reset!` it works and gives completions, and also squashed the commits.

Little related @arichiardi , there's a bug in lumo
lumo_infclojure
This can be reproduced in lumo like this
```
(let [ret (atom)] (lumo.repl/get-completions "red)" (fn [res] (reset! ret (map str res)))) @ret)
```
Otherwise I think @bbatsov can merge this now.

arichiardi:
LGTM

Issue #79, opened 2017-05-23
Eldoc error when running boot repl


AdamFrey:
## Expected behavior
Creates a usable inf-clojure repl using `boot repl`
## Actual behavior
An eldoc error is thrown constantly while typing in the repl
```
Debugger entered--Lisp error: (wrong-type-argument stringp nil)
  string-match("\\Ca*true\\Ca*" nil nil)
  #[257 "\301\302\303\304#)\207" [inhibit-changing-match-data "\\Ca*true\\Ca*" nil t string-match] 8 "\n\n(fn STRING)"](nil)
  inf-clojure--response-match-p("(js/global.hasOwnProperty \"$$LUMO_GLOBALS\")" #[257 "\301\302\303\304#)\207" [inhibit-changing-match-data "\\Ca*true\\Ca*" nil t string-match] 8 "\n\n(fn STRING)"] #<process inf-clojure<1>>)
  apply(inf-clojure--response-match-p ("(js/global.hasOwnProperty \"$$LUMO_GLOBALS\")" #[257 "\301\302\303\304#)\207" [inhibit-changing-match-data "\\Ca*true\\Ca*" nil t string-match] 8 "\n\n(fn STRING)"] #<process inf-clojure<1>>))
  inf-clojure--lumo-p(#<process inf-clojure<1>>)
  inf-clojure--detect-repl-type(#<process inf-clojure<1>>)
  inf-clojure--set-repl-type(#<process inf-clojure<1>>)
  inf-clojure-arglists-form()
  inf-clojure-arglists("nREPL")
  inf-clojure-eldoc-arglists(#("nREPL" 0 5 (fontified t face clojure-interop-method-face front-sticky (field inhibit-line-move-field-capture) rear-nonsticky t field output inhibit-line-move-field-capture t)))
  inf-clojure-eldoc()
  eldoc-print-current-symbol-info()
  #[0 "\204    \205\n\303>?\205\304 \207" [eldoc-mode global-eldoc-mode eldoc-documentation-function (nil ignore) eldoc-print-current-symbol-info] 2]()
  apply(#[0 "\204    \205\n\303>?\205\304 \207" [eldoc-mode global-eldoc-mode eldoc-documentation-function (nil ignore) eldoc-print-current-symbol-info] 2] nil)
[...]

PR #82, opened 2017-06-06
Fix for #79

arichiardi:

-----------------

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,26 @@ point. You can, however, change this behaviour by invoking such
 commands with a prefix argument. For instance: `C-u C-c C-v` will ask
 for the symbol you want to show the docstring for.
 
+#### Caveats
+
+Note that if you decide _NOT_ to use the socket repl, it is highly recommended
+you disable output coloring and/or readline facilities: `inf-clojure` does not
+filter out ASCII escape characters at the moment and will not behave correctly.
+
+You can disable coloring the following way for `boot`:
+
+```el
+((nil . ((inf-clojure-boot-cmd . "boot repl -C"))))
+```
+
+For leiningen, there are no command line switches and you need to add a custom [`project.clj` option](https://github.com/technomancy/leiningen/blob/master/sample.project.clj):
+
+```clojure
+...
+  :repl-options {:color false}
+...
+```
+
 ## Configuration options
 
 In the time-honoured Emacs tradition `inf-clojure`'s behaviour is extremely
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Bugs Fixed
 
 * [#77](https://github.com/clojure-emacs/inf-clojure/pull/77): Fix request "Eval expression:" if arglists return is `nil`.
+* [#79](https://github.com/clojure-emacs/inf-clojure/pull/82): Eldoc error when running boot repl.
 
 ## 2.0.0 (2017-05-01)
 
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -191,7 +191,7 @@ often connecting to a remote REPL process."
 
 (define-obsolete-variable-alias 'inf-clojure-program 'inf-clojure-lein-cmd "2.0.0")
 
-(defcustom inf-clojure-boot-cmd "boot repl"
+(defcustom inf-clojure-boot-cmd "boot repl -C"
   "The command used to start a Clojure REPL for Boot projects.
 
 Alternatively you can specify a TCP connection cons pair, instead


README.md
```diff
@@ -99,6 +99,26 @@ point. You can, however, change this behaviour by invoking such
 commands with a prefix argument. For instance: `C-u C-c C-v` will ask
 for the symbol you want to show the docstring for.
 
+#### *WARNING* - Standard REPLs
```
bbatsov: I think the title of this section is extremely misleading.

README.md
```diff
@@ -99,6 +99,26 @@ point. You can, however, change this behaviour by invoking such
...
+
+Note that if you decide _NOT_ to use the socket repl, it is highly recommended
+you disable output coloring and/or readline facilities: `inf-clojure` does not
+filter out ASCII escape characters at the moment and will not behave correctly.
```
bbatsov: Why filter them out? Why not simply apply them the same way we do in CIDER? It's quite trivial.

README.md
arichiardi: Suggestions?

README.md
arichiardi: We apply them visually, but the regex match is failing every time because of those. A solution is for our regexes to skip the ascii escape codes. Do you have a link for cider? If there is an alternative I am all for it.

hlolli:
Checked out this pull requests and this is what I'm getting
```
WARNING: No such namespace: complete.core, could not locate complete/core.cljs, complete/core.cljc, or Closure namespace "" at line 1 
             ⬆
WARNING: Use of undeclared Var complete.core/completions at line 1 
complete is not defined
	 (evalmachine.<anonymous>:1:1)
	 ContextifyScript.Script.runInThisContext (vm.cljs:44:33)
	 Object.runInThisContext (vm.cljs:116:38)
	 (Object.lumoEval)
	 lumo.repl.caching_node_eval (NO_SOURCE_FILE <embedded>:6133:194)
	 (NO_SOURCE_FILE <embedded>:5372:29)
	 x (NO_SOURCE_FILE <embedded>:5373:13)
	 Object.cljs.js.eval_str_STAR_ (NO_SOURCE_FILE <embedded>:5374:28)
	 Function.cljs.js.eval_str.cljs$core$IFn$_invoke$arity$5 (NO_SOURCE_FILE <embedded>:5377:508)
	 Object.lumo.repl.execute_text (NO_SOURCE_FILE <embedded>:6257:240)
```
and as I type `reduce` (which I fixed in my pull request)
```
Company: backend company-capf error "Wrong type argument: listp, ⬆" with args (candidates reduc)
Company: An error occurred in auto-begin
Company: backend company-capf error "Wrong type argument: listp, ⬆" with args (candidates reduce)
```

arichiardi:
Is this on Lumo? It means it recognizes it as Clojure Repl. Thanks for testing, very helpful.
-- 

hlolli:
On lumo yes (can be confirmed by looking at the stacktrace).

No problem!

arichiardi:
@hlolli how do you launch `lumo`? Because it does not happen here, I can see the eldoc. Lumo should be launched with `-d` because `inf-clojure` does not yet parses out `readline` and/or color escapes.

hlolli:
I did it like this
```
(defun lumo-start ()
  (interactive)
  (inf-clojure "cljs-boot lumo -K -d") 
  (with-current-buffer (buffer-name)
    (split-window-sensibly) 
    (previous-window)
    (previous-buffer)))
```

arichiardi:
@hlolli sorry about that, just double checking :) I can reproduce.

arichiardi:
You error above still happens when the first thing you do is `(redu` at a `lumo` REPL.

arichiardi:
@hlolli the code above should solve things. It would be great if you can apply #81 on top and check. I did that, and I see no errors in both lumo and normal repls.

README.md
bbatsov: Probably this should be in some caveats or faq section of the readme.

bbatsov: Or in configuration.

hlolli:
My changes for #81 are in your pull-request #82, so as expected, completions are working.

Eldoc is not on the other hand, just empty minibuffer.

But these three open pull-requests are now a soup.

arichiardi:
Oh ok I thought I took it off from mine, sorry about that! I guess @bbatsov can merge #81, than I can rebase on top of it.

README.md
arichiardi: The issue of speaking with the terminal is a big one. Comint hides some of the complexity but cannot eliminate it. Things like embedded `readline` and colors all change behaviors and I am so far from being an expert that I was even wondering whether console support should be dropped completely at some point (maybe by adopting another Emacs socket-only package). Sooner or later...

arichiardi: ...I guess it will happen. Cider does not handle this because of nRepl in between. In any case I can try to escape them, I saw it in some other plugin ...We'll see about the result.

arichiardi:
The eldoc error in Lumo is related to a `pr-str` added somehow to the command. Just discovered. I have a better patch for this, just need time to polish it up.

arichiardi:
Split this one in two (or more) while discovering and solving issues.

arichiardi:
@bbatsov so we are not merging this one? Why?

bbatsov:
My bad. Misinterpreted the comment.

arichiardi:
Ok cool lemme know 😀

PR #84, opened 2017-07-02
Disable pretty print when fetching lumo arglists

arichiardi:
Lumo has pretty print by default enabled. So this fix is necessary in order to avoid seeing newlines in eldoc.

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -703,7 +703,10 @@ If you are using REPL types, it will pickup the most approapriate
 (define-obsolete-variable-alias 'inf-clojure-arglist-command 'inf-clojure-arglists-form "2.0.0")
 
 (defcustom inf-clojure-arglists-form-lumo
-  "(pr-str (lumo.repl/get-arglists \"%s\"))"
+  "(let [old-value lumo.repl/*pprint-results*]
+     (set! lumo.repl/*pprint-results* false)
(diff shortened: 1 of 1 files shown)


inf-clojure.el
```diff
@@ -703,7 +703,7 @@ If you are using REPL types, it will pickup the most approapriate
 (define-obsolete-variable-alias 'inf-clojure-arglist-command 'inf-clojure-arglists-form "2.0.0")
 
 (defcustom inf-clojure-arglists-form-lumo
-  "(pr-str (lumo.repl/get-arglists \"%s\"))"
+  "(do (set! lumo.repl/*pprint-results* false) (lumo.repl/get-arglists \"%s\"))"
```
bbatsov: Why set this every time? I'd simply `let`-bind it around the other form.

arichiardi: Let bind does not work in cljs bootstrap. Still I guess I could do it once. There is a problem though...it is global so I might actually need to rethink this solution.

arichiardi:
Cannot think about a better solution. The only one would be to post process the output and make it single-lined after the evaluation. We alrerdy have all the pieces in place, just need to hook the post process function.

arichiardi:
Found a hack for this, works fine and does avoid changing the value.

bbatsov:
Looks pretty ugly, but I guess that's the best we can do for now.

PR #88, opened 2017-07-02
Introduce inf-clojure-log-activity

arichiardi:
Log commands and responses from/to the Inf-Clojure process. It can be enabled with `(setq inf-clojure-log-activity t)`. Vital for debugging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,6 +181,16 @@ directory and add this line to that file:
 jline.terminal=unsupported
 ```
 
+### Log process activity
+
+Standard Emacs debugging turns out to be difficult when an asynchronous process is involved. In this case try to enable logging:
+
+```el
+(setq inf-clojure-log-activity t)
+```
+
+This creates `.inf-clojure.log` in the process root for you to `tail -f` on.
+
 ## License
 
 Copyright © 2014-2017 Bozhidar Batsov and [contributors][].
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -946,7 +946,33 @@ prefix argument PROMPT-FOR-SYMBOL, it prompts for a symbol name."
 ;;;; Response parsing
 ;;;; ================
 
-(defvar inf-clojure--redirect-buffer-name " *Inf-Clojure Redirect Buffer*")
+(defvar inf-clojure--redirect-buffer-name " *Inf-Clojure Redirect Buffer*"
+  "The name of the buffer used for process output redirection.")
+
+(defvar inf-clojure--log-file-name ".inf-clojure.log"
+  "The name of the file used to log process activity.")
+
+(defvar inf-clojure-log-activity nil
+  "Log process activity?.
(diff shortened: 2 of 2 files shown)


arichiardi:
This one can be merged last as it will probably need to be rebased. Not vital and I prefer to work on this small one when the big ones are in.

inf-clojure.el
```diff
@@ -940,6 +940,25 @@ prefix argument PROMPT-FOR-SYMBOL, it prompts for a symbol name."
                (inf-clojure-symbol-at-point))))
     (comint-proc-query (inf-clojure-proc) (format (inf-clojure-var-source-form) var))))
 
+(defvar inf-clojure--dump-file-name ".inf-clojure.dump"
```
bbatsov: The more Emacsy way would be to dump debug output to some buffer, not a file (e.g. cider's message's buffer). I'd also add some commands like `inf-clojure-toggle-debug` (or something like this) and use `debug` instead of `dump` as the terminology for this feature.

arichiardi: Well I am not sure it is a debug mode...just dumping stuff. Anyways it is ok to me to call it debug.

About the file, I like that I can `tail -f` but I guess there is a way to save to disk or to tail once in a buffer without too much effort.

The tail is fundamental and saved me a couple of times already. I have had this feature for long time locally now.

arichiardi: Btw is there something like this in cider? Are we sure we want a buffer and not just a file? It is easier to implement (and maintain) I guess..

bbatsov: I was referring to the `nrepl-toggle-message-logging`. I don't like putting such logs in a file (by default), as refer to old logs extremely rarely. They are most useful for debug purposes right away. It's trivial to make a buffer scroll automatically when content is added to it.

README.md
```diff
@@ -181,6 +181,16 @@ directory and add this line to that file:
 jline.terminal=unsupported
 ```
 
+### Dump process activity
+
+Standard Emacs debugging turns out to be difficult when an asynchronous process is involved. In this case try to enable dumping:
+
+```el
+(setq inf-clojure-dump-activity t)
```
bbatsov: Let's at least call this `log-activity` instead of `dump-activity`.

README.md
```diff
@@ -181,6 +181,16 @@ directory and add this line to that file:
...
+```
+
+This creates `.inf-clojure.dump` in the process root for you to `tail -f` on.
```
bbatsov: And this could be a `.log`.

arichiardi:
Sure, this was just for me actually. I did some debugging today and thought I could use some improvement in the display :smile: If you want it in as it is I can change to `log` no problem.

bbatsov:
> I did some debugging today and thought I could use some improvement in the display 😄

Then do them. ;-)



> If you want it in as it is I can change to log no problem.

Yep.

arichiardi:
Done them :smile:

PR #93, opened 2017-07-10
Fix #89 and speed

### LINKED ISSUE #89: Slow response from inf-clojure (completions, arglists, ...)
Don't want to solution here, but it seems like `inf-clojure` got slower at responding in completions and eldoc/arglists.

It takes a while to see the completion even if it is very minimal, this must have been introduced with the comint buffer redirection. Probably an alternative implementation of `inf-clojure--process-response` can be attempted.

arichiardi:
Introduce double checking in the busy wait that detects if the inf-clojure process (the REPL) has completed and sent back everything. This removes the slowness that was triggered by having accept-process-output in the while body: we can take advantage now of its output to wait as well.

-----------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * [#79](https://github.com/clojure-emacs/inf-clojure/pull/82): Eldoc error when running boot repl.
 * [#83](https://github.com/clojure-emacs/inf-clojure/pull/85): No such namespace: complete.core in lumo REPL.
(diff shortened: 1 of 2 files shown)


bbatsov:
Just a small remark about the commit message style - when closing issues just prefix the commit title with `[Fix #NNNN] ...`.

bbatsov:
And great work overall! 👍

arichiardi:
Oh I always forget that right! Thanks for merging so quickly, `inf-clojure` is speedy again :smile:

PR #95, opened 2017-07-13
Calculate response boundaries explicitly

arichiardi:
Change the way the response is calculated, not by `search-forward` the buffer used for redirection but by parsing the raw response. Taking into consideration the optional regexps, a list is calculated with beginning, end and prompt position (the logic is now in inf-clojure--string-boundaries, which can be unit tested).

This implicitly solves a hidden bug in inf-clojure--process-response, which was not able to identify the prompt in some cases because of the many search-forward.

Finally, this new information is logged so that bugs can be detected more easily.

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -977,16 +977,27 @@ STRING if present."
                   'append
                   'no-annoying-write-file-in-minibuffer)))
 
+(defun inf-clojure--string-boundaries (string prompt &optional beg-regexp end-regexp)
+  "Calculate the STRING boundaries, including PROMPT.
+Return a list of positions (beginning end prompt).  If the
+optional BEG-REGEXP and END-REGEXP are present, the boundaries
+are going to match those."
+  (list (or (and beg-regexp (string-match beg-regexp string)) 0)
+        (or (and end-regexp (when (string-match end-regexp string)
+                              (match-end 0)))
+            (length string))
+        (or (string-match prompt string) (length string))))
+
 ;; Originally from:
 ;;   https://github.com/glycerine/lush2/blob/master/lush2/etc/lush.el#L287
-(defun inf-clojure--process-response (command process &optional beg-string end-string)
+(defun inf-clojure--process-response (command process &optional beg-regexp end-regexp)
   "Send COMMAND to PROCESS and return the response.
-Return the result of COMMAND starting with BEG-STRING and ending
-with END-STRING if non-nil.  If BEG-STRING is nil, the result
-string will start from (point) in the results buffer.  If
-END-STRING is nil, the result string will end at (point-max) in
-the results buffer.  It cuts out the output from and including
-the `inf-clojure-prompt`."
+Return the result of COMMAND, filtering it from BEG-REGEXP to the
+end of the matching END-REGEXP if non-nil.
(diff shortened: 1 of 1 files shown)


inf-clojure.el
```diff
@@ -1002,17 +1013,16 @@ the `inf-clojure-prompt`."
...
-                      (point-min)))
-             (end (or (when end-string
-                        (search-forward end-string nil t))
-                      (point-max)))
-             (prompt (when (search-forward inf-clojure-prompt nil t)
-                       (match-beginning 0)))
-             (buffer-string (buffer-substring-no-properties beg (or prompt end))))
-        (inf-clojure--log-string buffer-string "<-RES----")
-        buffer-string))))
+      (let* ((buffer-string (buffer-substring-no-properties (point-min) (point-max)))
```
bbatsov: What is `buffer-string` supposed to be here?

inf-clojure.el
```diff
@@ -1002,17 +1013,16 @@ the `inf-clojure-prompt`."
...
+             (boundaries (inf-clojure--string-boundaries buffer-string inf-clojure-prompt beg-regexp end-regexp))
+             (beg-pos (car boundaries))
+             (end-pos (car (cdr boundaries)))
+             (prompt-pos (car (cdr (cdr boundaries))))
+             (response-string (substring buffer-string beg-pos (min end-pos prompt-pos))))
+        (inf-clojure--log-string buffer-string "<-BUF----")
```
bbatsov: I think those markers you should would be understandable just to you. :-)

bbatsov: I also wonder who'd care about them most of the time.

bbatsov:
Parsing which string?

inf-clojure.el
arichiardi: uhm you mean you don't like the name? Naming is hard..

inf-clojure.el
arichiardi: It is a debugging feature, we can of course improve the naming and the markings, proposals are welcome!

arichiardi:
Improved description of the PR a bit.

inf-clojure.el
bbatsov: I really don't think this information would be useful to anyone 99% of the time, so I'm wary of logging it by default.

arichiardi: Keep in mind that it is logged only when `inf-clojure-log-activity` is `t`. It is not even a `defcustom` so you really are a dev/contributor if you are using this info. And the more info you have the more you debug nicely. Having said that, what would you remove? I am fine with putting the few lines I need when I debug.

bbatsov: Yeah, but that's not really process activity. It's just some random debug output that's completely useless unless you're debugging exactly this part of the code. I just think that this activity log should consistent and not a dumping ground for every debugging output one might decide to use. (and I usually debug with `edebug` so I'm not fond of dumping output like this in general)

arichiardi: I have tried with `edebug` to debug this kind of process output. No success. Maybe I was doing something wrong. In any case I can take it away. That is the core of the plugin and you need it like bread.

bbatsov: Well, OK. I still don't like it, but I'm also way to busy to discuss things of such little importance is great details.

Issue #98, opened 2017-07-26
Make variables safe


arichiardi:
## Expected behavior
Emacs should not complain for unsafety when configuring ` inf-clojure-boot-cmd`, for instance, in `.dir-locals.el`.
 
## Actual behavior
It does.

## Steps to reproduce the problem
Use this `.dir-locals.el` in your project and try to load a file, then change `inf-clojure-boot-cmd` and try to `revert-buffer`. It should not always ask for confirmation:

```
((nil . ((inf-clojure-boot-cmd . ("localhost" . 5555)))))
```

arichiardi:
Solved by #98 which carries a mistaken fix number. Ups.

PR #102, opened 2017-07-28
[Fix #89] Set defcustom safety

### LINKED ISSUE #89: Slow response from inf-clojure (completions, arglists, ...)

arichiardi:
Go over all the defcustom so that we set the correct `:safe` or `risky`
property.

Fix for #98.

Debatable and open for discussion:

The `inf-clojure-*-cmd`s will be considered risky if the value is a string (indicating a command line
execution) and safe is a cons cell containing a valid endpoint.

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -226,7 +226,7 @@ number (e.g. (\"localhost\" . 5555)).  That's useful if you're
 often connecting to a remote REPL process."
   :type '(choice (string)
                  (cons string integer))
-  :risky #'(stringp )
+  :risky #'stringp
   :safe #'inf-clojure--endpoint-p
   :package-version '(inf-clojure . "2.0.0"))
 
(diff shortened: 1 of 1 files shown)


inf-clojure.el
```diff
@@ -212,6 +226,8 @@ number (e.g. (\"localhost\" . 5555)).  That's useful if you're
 often connecting to a remote REPL process."
   :type '(choice (string)
                  (cons string integer))
+  :risky #'(stringp )
```
bbatsov: You make a typo here.

arichiardi: Oops thanks will fix

bbatsov:
You also have a typo in the first commit message (seems to me an `if` is missing there).

arichiardi:
Ok let me fix right away

arichiardi:
Done :+1:

PR #104, opened 2017-08-04
Protect against empty commands

arichiardi:
Improves the code so that empty strings are not sent to the process anymore. It also avoids trimming \t because they are actually used by some REPL.

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -256,15 +256,28 @@ It requires a REPL PROC for inspecting the correct type."
     inf-clojure-repl-type))
 
 (defun inf-clojure--single-linify (string)
-  "Convert a multi-line STRING in a single-line STRING."
-  (replace-regexp-in-string "[[:space:]\\|\n]+" " " string))
+  "Convert a multi-line STRING in a single-line STRING.
+It also reduces/adds redundant whitespace for readability.  Note
+that this function will transform the empty string in \" \" (it
+adds an empty space)."
+  (replace-regexp-in-string "[ \\|\n]+" " " string))
+
+(defun inf-clojure--trim-newline-right (string)
+  "Trim newlines (only) in STRING."
+  (if (string-match "\n+\\'" string)
+      (replace-match "" t t string)
+    string))
 
 (defun inf-clojure--sanitize-command (command)
   "Sanitize COMMAND for sending it to a process.
 An example of things that this function does is to add a final
-newline at the end of the form."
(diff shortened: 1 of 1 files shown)


inf-clojure.el
```diff
@@ -256,15 +256,28 @@ It requires a REPL PROC for inspecting the correct type."
...
 (defun inf-clojure--single-linify (string)
-  "Convert a multi-line STRING in a single-line STRING."
-  (replace-regexp-in-string "[[:space:]\\|\n]+" " " string))
+  "Convert a multi-line STRING in a single-line STRING.
+It also reduces/adds redundant whitespace for readability.  Note
+that this function will transform the empty string in \" \" (it
+adds an empty space)."
+  (replace-regexp-in-string "[ \\|\n]+" " " string))
+
+(defun inf-clojure--trim-newline-right (string)
```
bbatsov: That's a function in `subr-x.el`.

inf-clojure.el
```diff
@@ -256,15 +256,28 @@ It requires a REPL PROC for inspecting the correct type."
     inf-clojure-repl-type))
 
 (defun inf-clojure--single-linify (string)
-  "Convert a multi-line STRING in a single-line STRING."
-  (replace-regexp-in-string "[[:space:]\\|\n]+" " " string))
+  "Convert a multi-line STRING in a single-line STRING.
+It also reduces/adds redundant whitespace for readability.  Note
+that this function will transform the empty string in \" \" (it
+adds an empty space)."
+  (replace-regexp-in-string "[ \\|\n]+" " " string))
```
bbatsov: I guess this should also replace tabs.

inf-clojure.el
arichiardi: Well no, that is basically the point..the function does not want to replace tabs, because tabs can be sent to the Repl (for instance for completion)

inf-clojure.el
arichiardi: Same as above. Not replacing tabs on purpose.

inf-clojure.el
bbatsov: You might also want to mention this in the docstring.

inf-clojure.el
bbatsov: Same as above. ;-)

inf-clojure.el
arichiardi: Sure that is a good point.

inf-clojure.el
arichiardi: Yeah sure

PR #114, opened 2017-11-04
Introduce inf-clojure-project-type defcustom

arichiardi:
This patch adds a defcustom, inf-clojure-skip-project-detection, that when
non-nil skips the project detection so that only inf-clojure-generic-cmd will
be used.  This is very useful for projects that don't have standard layouts.

-----------------

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -189,6 +189,14 @@ number (e.g. (\"localhost\" . 5555))."
    (stringp (car x))
    (numberp (cdr x))))
 
+(defcustom inf-clojure-project-type nil
+  "Defines the project type.
+
+If this is `nil`, the project will be automatically detected."
+  :type 'string
+  :safe #'stringp
+  :package-version '(inf-clojure . "2.1.0"))
+
 (defcustom inf-clojure-lein-cmd "lein repl"
   "The command used to start a Clojure REPL for Leiningen projects.
 
@@ -511,10 +519,11 @@ Fallback to `default-directory.' if not within a project."
 
 (defun inf-clojure-project-type ()
   "Determine the type, either leiningen or boot of the current project."
-  (let ((default-directory (inf-clojure-project-root)))
-    (cond ((file-exists-p "project.clj") "lein")
-          ((file-exists-p "build.boot") "boot")
-          (t nil))))
+  (or inf-clojure-project-type
+      (let ((default-directory (inf-clojure-project-root)))
+        (cond ((file-exists-p "project.clj") "lein")
+              ((file-exists-p "build.boot") "boot")
+              (t "generic")))))
 
 (defun inf-clojure-cmd (project-type)
   "Determine the command `inf-clojure' needs to invoke for the PROJECT-TYPE."
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@
 * [#93](https://github.com/clojure-emacs/inf-clojure/pull/93): Slow response from inf-clojure (completions, arglists, ...).
 * [#101](https://github.com/clojure-emacs/inf-clojure/pull/101): `inf-clojure-set-ns` hangs Emacs.
 
+### New Features
+
+* [#114](https://github.com/clojure-emacs/inf-clojure/pull/114): Introduce `inf-clojure-project-type` defcustom.
+
 ## 2.0.1 (2017-05-18)
 
 ### Bugs Fixed
--- a/README.md
+++ b/README.md
@@ -74,6 +74,9 @@ interacting with it.
 `inf-clojure` has several custom variables which control the command
 used to start a REPL for particular project type - `inf-clojure-lein-cmd`,
 `inf-clojure-boot-cmd` and `inf-clojure-generic-cmd`.
+The `inf-clojure-project-type` can force a particular project type, skipping
+the project detection, which can be useful for projects that don't have
+standard layouts.
 
 By default all those variables are set to strings (e.g. `lein repl`).
 However, it is possible to use a cons pair like `("localhost" . 5555)`


bbatsov:
Frankly, I don't understand what exactly is the point of doing this change. You can always override the lein/boot command for a specific project, so if something is lein/boot and you need a different command to be used the option is always there for you. If it's not - you end up using the generic command anyways.

arichiardi:
I know very well that trick, using it all the time. However the scenario here is different: in lumo projects you don't need to have a `build.boot` or a `project.clj`...except that sometimes you want to resolve maven dependencies, which comes afterwards in most of my projects.

So all of a sudden you introduce the file (say `project.clj`) and boom, nothing works anymore because now `inf-clojure` detects and wants you to customize `inf-clojure-lein-cmd`. Ok I am going to do it. Say some time ago you realize that you want to switch to `build.boot`, because it is better (it really is faster for maven resolution). Boom again, now you have to customize `inf-clojure-boot-cmd`.

This back and forth happens in many scenarios (one other weird one is when you generate a project with `lein new` but you are not actually using `lein`, it is just for having a ready `project.clj`).

My rationale on this is that when I am in a `lumo` project I want to be able to decide what to do independently from the project type.

bbatsov:
I think what you really want to do is not to skip project detection but to be able to specify a generic project type via `.dir-locals.el`. That'd be the Emacs way of doing things.

arichiardi:
I am fine with that too, it still involves adding one more line in `.dir-locals.el`, same as this patch. Probably no difference. Will try that out cause if no patch necessary  it its better.

arichiardi:
Hello again @bbatsov, I have checked the code but it seems that there is [no `defcustom` for project type](https://github.com/clojure-emacs/inf-clojure/blob/master/inf-clojure.el#L542).

What would you suggest me to do here? Add one for project type? Then we'd need also a more flexible one for the command..(handle any possible commands for any possible project type?).

Seems like a bigger change.

arichiardi:
So to explain the problem better, in my folder I have both `project.clj` and `build.boot` and this is my `.dir-locals.el` with this patch.

```
((nil . ((eval . (add-hook 'clojure-mode-hook #'inf-clojure-minor-mode))
         (inf-clojure-repl-use-same-window . "different")
         (inf-clojure-skip-project-detection . t)
         (inf-clojure-generic-cmd . ("localhost" . 5044)))))
```

bbatsov:
> What would you suggest me to do here? Add one for project type? Then we'd need also a more flexible one for the command..(handle any possible commands for any possible project type?).

This just had to be checked by the `inf-clojure-project-type` command and take precedence over the existing detection logic. Doesn't seem like a big change to me.

arichiardi:
Got it, so you would prefer to create a `defcustom` containing a new project type.

What would the project type use as `-cmd`? I am not that elisp expert, can you create a `defvar` like `inf-clojure-${project-type}-cmd` from `.dir-locals.el`?

At the moment we have three `defcustom`: `inf-clojure-boot-cmd`, `inf-clojure-lein-cmd` and the catch all `inf-clojure-generic-cmd`.

arichiardi:
I see in `projectile` something like this, which makes total sense, but this changes things a bit:

```
(projectile-register-project-type 'npm '("package.json")
				  :compile "npm install"
				  :test "npm test"
				  :run "npm start"
				  :test-suffix ".spec")
```

arichiardi:
Maybe our detection can interact with `projectile` and see if there is a declared project type already and use that command declared in there. Projectile would need to handle the `:repl` tag though.

Fairly complicated I know, I still like my small hack better.

bbatsov:
> Got it, so you would prefer to create a defcustom containing a custom project type.

A `defvar` actually. It shouldn't be a custom project type. Just something that overrides the regular detection, that in your case you'd set to generic and also override the generic cmd. 

Seems to me that this solves your problem completely. 

> Maybe our detection can interact with projectile and see if there is a declared project type already and use that command declared in there. Projectile would need to handle the :repl tag though.

We can always add some integration with projectile as well - that'd be pretty simple. But this should be something optional, as we don't want to force Projectile on everyone.

arichiardi:
Isn't a defvar that override the logic exactly what I am doing? Maybe I don't understand your solution here.

arichiardi:
The ` inf-clojure-skip-project-detection` overrides code detection so that we always take the generic command.

bbatsov:
Yeah, but you're making something very specific (you can just make something generic) and it should be able to force whatever project type you want to. That's what started all of my comments. I don't like very specific options, normally I'd make something as generic as possible.

inf-clojure.el
```diff
@@ -189,6 +189,15 @@ number (e.g. (\"localhost\" . 5555))."
    (stringp (car x))
    (numberp (cdr x))))
 
+(defcustom inf-clojure-skip-project-detection nil
```
bbatsov: Basically this should be `inf-clojure-project-type` and the docstring should say it's purpose is to override the default project type resolution via `.dir_locals.el`.

inf-clojure.el
```diff
@@ -510,10 +519,11 @@ Fallback to `default-directory.' if not within a project."
 
 (defun inf-clojure-project-type ()
   "Determine the type, either leiningen or boot of the current project."
-  (let ((default-directory (inf-clojure-project-root)))
-    (cond ((file-exists-p "project.clj") "lein")
-          ((file-exists-p "build.boot") "boot")
-          (t nil))))
```
bbatsov: We should be returning `"generic"` if the project type can't be determined, so we can leverage the `defcustom` - if it's nil it does nothing, otherwise it takes precedence over this.

bbatsov:
I've explained in the code exactly what I mean. Hopefully now you understand me.

arichiardi:
Thanks should be clear now, will change this patch.

arichiardi:
@bbatsov done, can you review please. I want to be sure I got it right.

inf-clojure.el
```diff
@@ -511,10 +519,11 @@ Fallback to `default-directory.' if not within a project."
...
+  (when (null inf-clojure-project-type)
```
bbatsov: Why don't you use `or` here?

arichiardi: will do

arichiardi: No wait, the code should run if this is `nil`. So basically something else than `nil` will not trigger the detection.

bbatsov: Well, that's what `or` will do - `(or type (detect-type))`. That seems pretty clear to me.

arichiardi: You are right, I was confused for a moment :smile:

PR #115, opened 2017-11-17
Add :safe clause for inf-clojure-repl-use-same-window

arichiardi:

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -367,6 +367,7 @@ This should usually be a combination of `inf-clojure-prompt' and
   "Controls whether to display the REPL buffer in the current window or not."
   :type '(choice (const :tag "same" t)
                  (const :tag "different" nil))
+  :safe #'booleanp
   :package-version '(inf-clojure . "2.0.0"))
 
 (defvar inf-clojure-buffer nil


inf-clojure.el
```diff
@@ -367,6 +367,7 @@ This should usually be a combination of `inf-clojure-prompt' and
   "Controls whether to display the REPL buffer in the current window or not."
   :type '(choice (const :tag "same" t)
                  (const :tag "different" nil))
+  :safe (lambda (s) (or (string-equal "same" s) (string-equal "different" s)))
```
bbatsov: The actual values are `t` and `nil`, not "same" and "different". A guess this should just be a `boolean?`.

arichiardi:
Oh, well it is working this way, emacs does no show it as unsafe, maybe becuaes a string is `t`?

bbatsov:
I think you should just use `booleanp` here.

bbatsov:
Here's an example - https://github.com/clojure-emacs/cider/blob/976ffdbb57bf1a394c4705bdf0ed773c054c63bc/cider-eldoc.el#L85

arichiardi:
Will do that

PR #119, opened 2017-12-30
 [Fix #116] Set inf-clojure-buffer REPL type on detect

### LINKED ISSUE #116: Repl buffer sometimes has different inf-clojure-repl-type
Probably due to some async interleaving sometimes I see the REPL buffer having a different repl type compared with the code buffer.

This means completion is failing with very strange errors.

arichiardi:
Now the code sets the `inf-clojure-repl-type` buffer local var in the `inf-clojure-buffer` when a REPL type is detected. This solves the weird errors that were happening when working within the REPL buffer because `inf-clojure-repl-type` is nil.

-----------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * [#83](https://github.com/clojure-emacs/inf-clojure/pull/85): No such namespace: complete.core in lumo REPL.
 * [#93](https://github.com/clojure-emacs/inf-clojure/pull/93): Slow response from inf-clojure (completions, arglists, ...).
 * [#101](https://github.com/clojure-emacs/inf-clojure/pull/101): `inf-clojure-set-ns` hangs Emacs.
+* [#119](https://github.com/clojure-emacs/inf-clojure/pull/119): Set inf-clojure-buffer REPL type on detect.
 * [#120](https://github.com/clojure-emacs/inf-clojure/pull/120): Send REPL string always, even if empty.
 
 ### New Features
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -292,9 +292,10 @@ See http://blog.jorgenschaefer.de/2014/05/race-conditions-in-emacs-process-filte
 (defun inf-clojure--set-repl-type (proc)
   "Set the REPL type if has not already been set.
 It requires a REPL PROC for inspecting the correct type."
-  (if (not inf-clojure-repl-type)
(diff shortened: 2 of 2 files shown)


arichiardi:
This was indeed easy, style can be improved because `setq` returns the value so everything can be threaded but it looks okay this way too.

inf-clojure.el
```diff
@@ -293,7 +293,10 @@ See http://blog.jorgenschaefer.de/2014/05/race-conditions-in-emacs-process-filte
   "Set the REPL type if has not already been set.
 It requires a REPL PROC for inspecting the correct type."
   (if (not inf-clojure-repl-type)
-      (setq inf-clojure-repl-type (inf-clojure--detect-repl-type proc))
+      (let ((detected-type (inf-clojure--detect-repl-type proc)))
+        (with-current-buffer inf-clojure-buffer
+          (setq inf-clojure-repl-type detected-type))
+        (setq inf-clojure-repl-type detected-type))
```
bbatsov: Why do we need to set the REPL type in the source buffer? We can always infer this from the REPL buffer for the project.

arichiardi: Uhm are there any implications in terms of performance in getting the buffer and from it the value for every operation? Remember this is done on each key press (for completion). If not I can change things, but this patch was definitely easier.

I think also that removing the buffer local would work but I was thinking that at some point we should have more than one repl (per project) so I don't know how it work then.

arichiardi:
I updated the code and also notice that we have a [`defvar-local`](https://github.com/arichiardi/inf-clojure/blob/set-repl-buffer-repl-type/inf-clojure.el#L274) that can be manually set. As long as it is not `nil` detection will be skipped.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 * [#83](https://github.com/clojure-emacs/inf-clojure/pull/85): No such namespace: complete.core in lumo REPL.
 * [#93](https://github.com/clojure-emacs/inf-clojure/pull/93): Slow response from inf-clojure (completions, arglists, ...).
 * [#101](https://github.com/clojure-emacs/inf-clojure/pull/101): `inf-clojure-set-ns` hangs Emacs.
+* [#119](https://github.com/clojure-emacs/inf-clojure/pull/119): Send REPL string always, even if empty.
```
bbatsov: I think you put to wrong changelog entry here.

arichiardi: Yep I renamed things and forgot it, will do

PR #122, opened 2017-12-31
Introduce inf-clojure-completions-fn defcustom

arichiardi:
Now the user can provide a custom parsing function for completion and therefore
has complete (pun intended) freedom in what to use for it. Some could use
compliment completion for instance or even use directly what cider provides.
The defcustom defaults to inf-clojure-list-completions, which can only parse
candidates coming as a Lisp list of strings.

-----------------

Thanks!

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -1210,7 +1210,7 @@ prefix argument PROMPT-FOR-NS, it prompts for a namespace name."
 
 (defun inf-clojure-set-ns (prompt-for-ns)
   "Set the ns of the inferior Clojure process to NS.
-See variable `inf-clojure-set-ns-form`.  It defaults to the ns of
+See variable `inf-clojure-set-ns-form'.  It defaults to the ns of
 the current buffer.  When invoked with a prefix argument
 PROMPT-FOR-NS, it prompts for a namespace name."
   (interactive "P")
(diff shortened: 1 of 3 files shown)


inf-clojure.el
```diff
@@ -1254,15 +1254,43 @@ See variable `inf-clojure-buffer'."
   "Return DATA if and only if it is a list."
   (when (listp data) data))
 
-(defun inf-clojure-completions (expr)
-  "Return a list of completions for the Clojure expression starting with EXPR."
+(defun inf-clojure-list-completions (expr)
+  "Return completions for the Clojure expression starting with EXPR.
+
+Its only ability is to parse a list of candidate strings, every
+other EXPR will be discarded and nil will be returned."
```
bbatsov: Isn't EXPR only the completion prefix? Probably you didn't mean to reference it here. On a related note - we might pick a better name for this param.

inf-clojure.el
```diff
@@ -1254,15 +1254,43 @@ See variable `inf-clojure-buffer'."
...
+candidates (e.g.: (\"def\" \"defn\")) but more complex libraries
+like `alexander-yakushev/compliment' can return other things like
+edn.
+
+The expected return depends on the mode that you use for
+completion: `company-mode' for instance allows metadata along
+with the candidates completion. See \\[completion-table-dynamic]
+for more details.
+
+The default value is the `inf-clojure-list-completions' function,
```
bbatsov: As the the default function does more than parsing the results (it's also invokes the Clojure completion code) this should probably the adjusted.

inf-clojure.el
```diff
@@ -1254,15 +1254,43 @@ See variable `inf-clojure-buffer'."
...
+Its only ability is to parse a list of candidate strings, every
+other EXPR will be discarded and nil will be returned."
   (when (not (string-blank-p expr))
     (thread-first
         (format (inf-clojure-completion-form) (substring-no-properties expr))
       (inf-clojure--process-response (inf-clojure-proc) "(" ")")
       (inf-clojure--read-or-nil)
       (inf-clojure--list-or-nil))))
 
+(defcustom inf-clojure-completions 'inf-clojure-list-completions
```
bbatsov: Won't `inf-clojure-completions-fn` (or something like this) be a better name?

inf-clojure.el
```diff
@@ -1254,15 +1254,43 @@ See variable `inf-clojure-buffer'."
   "Return DATA if and only if it is a list."
   (when (listp data) data))
 
-(defun inf-clojure-completions (expr)
-  "Return a list of completions for the Clojure expression starting with EXPR."
+(defun inf-clojure-list-completions (expr)
```
bbatsov: I don't like this name, as this function also invokes the code which generates the list it eventually processes. Not sure we won't to decouple only the part that actually processes the results. The way I see it completion has 3 stages:
* decide what's the Clojure code that will do the completion (e.g. we can set this per project or have some abilities check - try if compliment is present, fallback to `core.complete` if this fails. Probably the first option would be easier and a better initial approach to this. )
* invoke that code and get some result from it
* decide how to process the results

inf-clojure.el
arichiardi: Open to suggestions :)

inf-clojure.el
arichiardi: I like this better yes

inf-clojure.el
arichiardi: Yes I think I will take the send response out of the default function. I did not notice that `inf-clojure-process-response` was there. In my mind only the string should be passed down to the custom parsing function, as you said above.

inf-clojure.el
arichiardi: Will refactor

PR #124, opened 2018-01-05
WIP - Introduce inf-clojure-compliment completions

arichiardi:
Given compliment on the classpath, the inf-clojure-compliment can be required
in order to customize the right defcustoms that are necessary for it.
Completions are only annotated and in order to customize this process an
inf-clojure hacker can use inf-clojure-completions-annotation-fn.

-----------------

### DIFF
--- a/inf-clojure-compliment.el
+++ b/inf-clojure-compliment.el
@@ -1,0 +1,112 @@
+;;; inf-clojure-compliment.el --- Compliment completion file for Inf-Clojure -*- lexical-binding: t; -*-
+;;
+;; Copyright © 2014-2018 Bozhidar Batsov
+
+;; Authors: Bozhidar Batsov <bozhidar@batsov.com>
+;;       Andrea Richiardi <a.richiardi.work@gmail.com>
+;; URL: http://github.com/clojure-emacs/inf-clojure
+;; Keywords: processes, clojure
+;; Version: 2.1.0
+;; Package-Requires: ((emacs "24.4") (edn "1.1.2"))
+
+;; This file is part of GNU Emacs.
+
+;; GNU Emacs is free software: you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; GNU Emacs is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with GNU Emacs.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+;;
+;; Code completion using alexander-yakushev/compliment.
+
+;;; Code:
+
+(require 'edn)
+
+(defconst inf-clojure-compliment--annotations-alist
+  '((:class "c")
+    (:field "fi")
+    (:function "f")
+    (:import "i")
+    (:keyword "k")
+    (:local "l")
+    (:macro "m")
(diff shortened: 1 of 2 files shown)


arichiardi:
@bbatsov please review this, I am sure there are many things to improve. After that I will try to add docs.

inf-clojure-compliment.el
```diff
@@ -0,0 +1,112 @@
...
+every other EXPR will be discarded and nil will be returned."
+  (mapcar #'inf-clojure-compliment--parse-candidate
+          (thread-first
+              response-str
+            (inf-clojure--some)
+            (edn-read))))
+
+(setq inf-clojure-completions-fn #'inf-clojure-compliment--completions)
+(setq inf-clojure-completions-annotation-fn #'inf-clojure-compliment--annotate-symbol)
+(setq inf-clojure-completion-form "(do (require '[compliment.core :as comp]) (compliment.core/completions \"%s\"))")
```
bbatsov: You don't use the `comp` alias.

bbatsov: I also wonder if this should be a no-op or raise some error if the `require` fails.

bbatsov:
Yeah, this looks reasonable, but in it's present form it should a separate package on MELPA, which I presume if your idea.

inf-clojure-compliment.el
arichiardi: Currently everything that fails emacs'  `read` is considered an error and returns `nil`.
There is a function, I think `inf-clojure--read-or-nil`. I actually believe I should still dump errors on a file or something and not just discard them.

arichiardi:
A package only for this seems a bit overkoll no?

arichiardi:
A package only for this seems a bit overkoll no?

arichiardi:
I am thinking a user could either include it in her init file or in `.dir-locals.el`. Possibly the latter if your projects are not only clj projects. Inf-clolure is more hacky anyways. If you have some other idea I am open to it.

bbatsov:
> A package only for this seems a bit overkill no?

The problem is that the main package doesn't depend on `edn.el`. Perhaps this was just a mistake you made. That's why I asked. Perhaps even a separate file is an overkill then.

arichiardi:
Yeah well I put it together so quickly :wink: we could depend on it I guess?

bbatsov:
Yeah, I'm fine with that.

arichiardi:
I am not yet happy with the `require` there though. I am thinking that we should have a step for initializing the REPL with custom code. At the moment the only init code that we have is `find-ns` that reports which is the REPL type.
However this can be a highly customizable thing. Say the `require` of `compliment` goes through fine, I might want to return a key like `{:completion-engine :compliment}`. This can be merged with `{:repl-type :clojure}` for instance, which `inf-clojure` itself needs.
Each "init form" can return something and we could accumulate the returns into a map. This map is our REPL hello. In the future this REPL hello can come from `unrepl` for instance.

bbatsov:
Yeah, that would certainly be a better approach, and it become quite handy down the road. Good idea!

bbatsov:
Btw, I just realized you can just inject the compliment dep when starting the REPL via lein/boot/clj via a command-line option and make this the default completion. That'd be pretty cool I think.

This came to my mind after I replied to your message on Slack when I said `orchard` could be pretty useful to `inf-clojure` as well - e.g. a much better find-source, apropos, etc.

arichiardi:
Yep you can do that `boot -D compliment:0.3.5` works very well. The "detection" of the REPL capabilities is another cattle of fish and is akin to `unrepl` `hello` command. In our case it can happen any time though (in our case on the first interaction). I found this better because you could inject dependencies at runtime and "have a new feature" in your REPL.

Anyways did not have time to continue this PR for now...but it is an exciting one ;)

bbatsov:
I'm closing this PR due to no recent activity. Feel free to re-open it if you ever come back to it.

PR #125, opened 2018-01-12
Avoid throwing an error for frequent operations like completion

arichiardi:
This patch introduces reduces `inf-clojure-proc`'s responsabilities to only one:
return the processs. The new `inf-clojure--proc-or-error` will do both returning
and erroring.
This is necessary because while it is good to communicate an error on "eval"
operations, it is not good to have it on "under the hood" and very frequent
ones like completion and get arglists.

-----------------

Thanks!

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -363,11 +363,11 @@ Clojure to load that file."
   :safe #'stringp
   :package-version '(inf-clojure . "2.0.0"))
 
-(defun inf-clojure-load-form ()
-  "Return the form to query inferior Clojure for a var's documentation.
+(defun inf-clojure-load-form (proc)
+  "Return the form to query the Inf-Clojure PROC for var's documentation.
 If you are using REPL types, it will pickup the most appropriate
 `inf-clojure-var-doc-form` variant."
-  (pcase (inf-clojure--set-repl-type (inf-clojure-proc))
+  (pcase (inf-clojure--set-repl-type proc)
     (`lumo inf-clojure-load-form-lumo)
     (`planck inf-clojure-load-form-planck)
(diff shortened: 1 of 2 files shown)


arichiardi:
I also want to pass `proc` through all the `inf-clojure-*-form` functions because it seems to me a bit ugly and brittle that they get `proc` independently. It will take a little bit more time though so I am leaving this WIP (and try it anyways).

inf-clojure.el
```diff
@@ -1204,52 +1207,60 @@ prefix argument PROMPT-FOR-SYMBOL, it prompts for a symbol name."
...
-  (let ((proc (get-buffer-process (if (derived-mode-p 'inf-clojure-mode)
-                                      (current-buffer)
-                                    inf-clojure-buffer))))
-    (or proc
-        (error "No Clojure subprocess; see variable `inf-clojure-buffer'"))))
+  (get-buffer-process (if (derived-mode-p 'inf-clojure-mode)
+                          (current-buffer)
+                        inf-clojure-buffer)))
+
+(defun inf-clojure--proc-or-error ()
```
bbatsov: I think a more common approach is to have just one function and some param controlling wether to raise errors or return `nil`. That's what we do in CIDER for instance (and I've seen in many other packages).

arichiardi: The function won't be pure anymore though

bbatsov: In Emacs that's totally fine. :-)

bbatsov:
The changes look good, but your branch has to be rebased on top of the current `master` branch due to merge conflicts.

arichiardi:
I am going to do it now

PR #126, opened 2018-01-13
Derive keymap from prog-mode

arichiardi:
-----------------

It is the right thing to do so that the keymap shares previously loaded
prog-mode bindings.


bbatsov:
Only `major-mode` are supposed to derive from `prog-mode`, so that change doesn't make sense. `clojure-mode` already derives from `prog-mode`, as it's supposed to.

PR #127, opened 2018-01-13
Fix and reword docstring

arichiardi:
I went through a checkdoc session and fixed what was easiest to fix.  Not
everything, but it is a good start.

-----------------

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -392,7 +392,7 @@ This should usually be a combination of `inf-clojure-prompt' and
   :package-version '(inf-clojure . "2.0.0"))
 
 (defvar inf-clojure-buffer nil
-  "The current inf-clojure process buffer.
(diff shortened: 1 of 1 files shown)


inf-clojure.el
```diff
@@ -392,7 +392,7 @@ This should usually be a combination of `inf-clojure-prompt' and
   :package-version '(inf-clojure . "2.0.0"))
 
 (defvar inf-clojure-buffer nil
-  "The current inf-clojure process buffer.
+  "The current ‘inf-clojure’ process buffer.
```
bbatsov: It should be `\`inf-clojure'`.

bbatsov: Yeah, I just think you didn't put the right quotes around it. Only the first one should be a backquote.

arichiardi:
It looks like the ```inf-clojure'` syntax is fine.

PR #129, opened 2018-01-22
Improve the completion bounds detection (now with keywords)

arichiardi:
It was missing, but it is actually used for identifying completions bounds.

-----------------

### DIFF
--- a/test/inf-clojure-tests.el
+++ b/test/inf-clojure-tests.el
@@ -1,0 +1,114 @@
+;;; inf-clojure-tests.el --- Tests for Inf-Clojure -*- lexical-binding: t; -*-
+;;
+;; Copyright © 2014-2018 Bozhidar Batsov
+
+;; Authors: Bozhidar Batsov <bozhidar@batsov.com>
+;;          Andrea Richiardi <a.richiardi.work@gmail.com>
+;; URL: http://github.com/clojure-emacs/inf-clojure
+;; Keywords: processes, clojure
+;; Package-Requires: ((emacs "24.4") (clojure-mode "5.3"))
+
+;; This file is part of GNU Emacs.
+
+;; GNU Emacs is free software: you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; GNU Emacs is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with GNU Emacs.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+;;
+;; Code completion using alexander-yakushev/compliment.
+
+;;; Code:
+
+;; Tests for inf-clojure.el
+
+(message "Running tests on Emacs %s" emacs-version)
+
+(require 'buttercup)
+(require 'assess)
+(require 'inf-clojure)
+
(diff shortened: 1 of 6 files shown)


arichiardi:
This needs more work as other things are failing, like completing `^:private`.

arichiardi:
The two changes are running fine locally, the problem is the tests are failing for some reason.

arichiardi:
I think I need some help with the tests because they are not very nice. For instance you cannot understand what is wrong when something fails.

bbatsov:
Btw, you might want to also consider `buttercup` (https://github.com/jorgenschaefer/emacs-buttercup) for the tests. I find it much nicer to use than ert. I use it on a few projects and I like it quite a lot. I think the only thing ert has going is that it's built-in.

arichiardi:
I have checked buttercup thoroughly and stamped upon https://github.com/jorgenschaefer/emacs-buttercup/pull/37.

It seems things are in motion, slowly, but `assess` is probably the way for me to go.

It should end up in core and probably at some point will have BDD testing in it: https://github.com/phillord/assess/blob/master/README.md#roadmap

I so like the helper functions for buffer testing, which is basically what I am doing here, so if you don't mind I will try that first.

test/inf-clojure-tests.el
```diff
@@ -0,0 +1,86 @@
...
+          (it "computes bounds for @deref when point is after it"
+              (ict-with-assess-buffers
+               ((a (progn
+                     (insert "@deref")
+                     (goto-char 7))))
+               (with-current-buffer a
+                 (expect (ict-bounds-string (inf-clojure-completion-bounds-of-expr-at-point))
+                         :to-equal "deref"))))
+
+          (it "computes bounds for ^:keyword when point is after it"
```
arichiardi: This is failing still, but I moved to `buttercup` and `assess`

bbatsov: What is it actually matching?

test/inf-clojure-tests.el
```diff
@@ -0,0 +1,86 @@
...
+               ((a (insert "plain-text")))
+               (with-current-buffer a
+                 (expect (ict-bounds-string (inf-clojure-completion-bounds-of-expr-at-point))
+                         :to-equal "plain-text"))))
+
+          (it "computes bounds for @deref when point is after it"
+              (ict-with-assess-buffers
+               ((a (progn
+                     (insert "@deref")
+                     (goto-char 7))))
```
bbatsov: Isn't your cursor right after the word when you do `insert` anyways?

arichiardi: Well, now I know :wink:

test/inf-clojure-tests.el
arichiardi: it should compute `:keyword` without the hat

arichiardi: Now it is returning `nil`

bbatsov: And what happens when you try this interactively?

arichiardi: I tried not in a test, in normal buffer it was working

arichiardi: So if you open an `inf-clojure-minor-mode` buffer and put this in it:

```
^:pr|
```

In `ielm` you'll see:

```
ELISP> (inf-clojure-completion-bounds-of-expr-at-point)
(2 . 5)
```

arichiardi:
I ended up improving the regex (big dragons in elisp regexes!) and adding keyword detection so that now tests are passing with flying green colors.

bbatsov:
There's a conflict with your other PRs, but overall the changes look good.

arichiardi:
@bbatsov Can't wait to have tests in :smile:

bbatsov:
Your wait is over. ;-)

arichiardi:
Thank you @bbatsov!

PR #132, opened 2018-01-31
Introduce inf-clojure-reload

arichiardi:
It will evaluate (require 'ns :reload) or (require 'ns :reload-all) at the
REPL, depending on the arguments passed in.

  - `C-u M-x inf-clojure-reload`: prompts for a namespace name.
  - `M-- M-x inf-clojure-reload`: executes (require ... :reload-all).
  - `M-- C-u M-x inf-clojure-reload`: reloads all AND prompts."

This works only from source buffers at the moment (like all the other namespace commands).

-----------------

Thanks!

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -131,6 +131,7 @@ mode.  Default is whitespace followed by 0 or 1 single-letter colon-keyword
     (define-key map "\C-c\C-c" #'inf-clojure-eval-defun)     ; SLIME/CIDER style
     (define-key map "\C-c\C-b" #'inf-clojure-eval-buffer)
     (define-key map "\C-c\C-r" #'inf-clojure-eval-region)
+    (define-key map "\C-c\M-r" #'inf-clojure-reload)
     (define-key map "\C-c\C-n" #'inf-clojure-eval-form-and-next)
     (define-key map "\C-c\C-z" #'inf-clojure-switch-to-repl)
     (define-key map "\C-c\C-i" #'inf-clojure-show-ns-vars)
@@ -152,6 +153,7 @@ mode.  Default is whitespace followed by 0 or 1 single-letter colon-keyword
         ["Eval buffer" inf-clojure-eval-buffer t]
         "--"
         ["Load file..." inf-clojure-load-file t]
+        ["Reload file... " inf-clojure-reload t]
         "--"
         ["Switch to REPL" inf-clojure-switch-to-repl t]
         ["Set REPL ns" inf-clojure-set-ns t]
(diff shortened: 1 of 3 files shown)


inf-clojure.el
```diff
@@ -702,6 +746,35 @@ is present it will be used instead of the current file."
...
+(defun inf-clojure-reload (arg)
+  "Send a query to the inferior Clojure for reloading the namespace.
+See variable `inf-clojure-reload-form' and
+`inf-clojure-reload-all-form'.
+
+The prefix argument ARG can change the behavior of the command:
+
+  - C-u M-x `inf-clojure-reload': prompts for a namespace name.
+  - M-- M-x `inf-clojure-reload': inverts the default reload command
+  - M-- C-u M-x `inf-clojure-reload': inverts default AND prompts for a namespace name."
```
arichiardi: There is probably better syntax for this...

bbatsov: http://ergoemacs.org/emacs/inline_doc.html

Anyways, this will do for now.

inf-clojure.el
```diff
@@ -702,6 +746,35 @@ is present it will be used instead of the current file."
...
+  (interactive "P")
+  (let* ((proc (inf-clojure-proc))
+         (invert? (or (equal current-prefix-arg "-") (equal current-prefix-arg '(-4))))
```
bbatsov: `arg` is your prefix argument, so no need to use `current-prefix-argument`.

inf-clojure.el
```diff
@@ -702,6 +746,35 @@ is present it will be used instead of the current file."
...
+         (prompt? (or (equal current-prefix-arg '(4)) (equal current-prefix-arg '(-4))))
```
bbatsov: Remember we don't use `?` in Emacs - it's `p` there.

inf-clojure.el
```diff
@@ -702,6 +746,35 @@ is present it will be used instead of the current file."
     (when switch-to-repl
       (inf-clojure-switch-to-repl t))))
 
+(defcustom inf-clojure-reload-default t
```
bbatsov: Why not `'reload` and `'reload-all`? Boolean predicate for this seems pretty confusing to me. It's better to use a symbol.

inf-clojure.el
bbatsov: Btw, having 3 ways to invoke something becomes pretty confusing very fast - normally people do just single and double prefix arguments. I think the default inversion is a bit too much. I'd just drop this and the config and do two commands probably.

inf-clojure.el
arichiardi: Right!

inf-clojure.el
arichiardi: It's just that they are so similar that you usually think of it as same. That is the reason why the same key binding is used.

The inversion of the default can be indeed dropped and we can just always default for `reload`. With `M--` `reload-all`.
I have used this for a week now and found it good enough.

inf-clojure.el
arichiardi: Will change

inf-clojure.el
arichiardi: I will probably drop this

arichiardi:
@bbatsov I have just noticed that there is a similar key binding in `clojure-mode` so please don't merge this I want to change it to `C-c M-r`.

inf-clojure.el
```diff
@@ -131,6 +131,7 @@ mode.  Default is whitespace followed by 0 or 1 single-letter colon-keyword
     (define-key map "\C-c\C-c" #'inf-clojure-eval-defun)     ; SLIME/CIDER style
     (define-key map "\C-c\C-b" #'inf-clojure-eval-buffer)
     (define-key map "\C-c\C-r" #'inf-clojure-eval-region)
+    (define-key map "\C-cM-r" #'inf-clojure-reload)
```
bbatsov: I think you missed a `\` here.

arichiardi: Yeah will fix today thanks!

PR #138, opened 2018-02-24
Set buffer local comint-input-sender-no-newline

arichiardi:
Now that the sanitation code appends a final newline to the command we need to
tell comint not to add one.

Together with https://github.com/clojure-emacs/inf-clojure/pull/137, it should
bring things back to normal and definitely more correct.

-----------------

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -177,6 +177,9 @@ The following commands are available:
 
 \\{inf-clojure-minor-mode-map}"
   :lighter "" :keymap inf-clojure-minor-mode-map
+  ;; it is not originally buffer local
+  ;; https://github.com/emacs-mirror/emacs/blob/master/lisp/comint.el#L415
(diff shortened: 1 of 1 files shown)


arichiardi:
So for posterity, this solution does not actually work.

Because of [process-send-eof](https://github.com/emacs-mirror/emacs/blob/master/lisp/comint.el#L2310) called at the end of `comint-simple-send`, which I have no idea of what is intended to achieve, an error is throw saying that there is no process associated with the buffer we execute in, which is a source buffer (only `*inf-clojure*` has a process associated with it).

So either we set a process for our source files, losing the flexibility of having many `*inf-clojure*` per source file, or we use something else for sending commands.

A super low-level [comint-send-string](https://github.com/emacs-mirror/emacs/blob/master/lisp/comint.el#L2307) is used there and I am going to try that one in another PR.

PR #139, opened 2018-02-24
Use comint-send-string

arichiardi:
This one is solving the weird duplicating prompt issue. After the sanitation
patch we were sending newline twice.  Using comint-send-string because it is
more low-level and does not modify the input string, which we are building
correctly already (hopefully).

See also https://github.com/clojure-emacs/inf-clojure/pull/138.

-----------------

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -342,7 +342,7 @@ customizations."
   (inf-clojure--set-repl-type proc)
   (let ((sanitized (inf-clojure--sanitize-command string)))
     (inf-clojure--log-string sanitized "----CMD->")
-    (comint-simple-send proc sanitized)))
+    (comint-send-string proc sanitized)))
 
 (defcustom inf-clojure-load-form "(clojure.core/load-file \"%s\")"
   "Format-string for building a Clojure expression to load a file.


arichiardi:
This needs to be merged before the other one in order to avoid conflicts.

arichiardi:
No will have conflicts, tried locally.

j0ni:
Hmm, I tried this tweak locally and though the immediate result is positive (only one prompt was output after evaluating a simple form), after about 20 seconds, during which I moved around two or three clj file buffers but didn't evaluate anything, three more prompts were printed out in the `*inf-clojure*` buffer.

No idea if it's related to the change, maybe this is another issue.

j0ni:
(now I can't reproduce it :man_shrugging: )

arichiardi:
This goes in combination with the other PR, thanks for trying Btw!

j0ni:
you're welcome - what else am I going to do on the weekend? :stuck_out_tongue:

By "other PR", do you mean #137 ?

arichiardi:
Yep, well I spent quite a bit of my night yesterday to try out solutions and this seems the most reliable. On which repl are you trying?

j0ni:
I was using nREPL started via a leiningen project.

I'm out right now but later I'll try it with a socket repl.

arichiardi:
Ok let me know, folks trying other repls are invaluable given that I usually work with `lumo`

bbatsov:
You'll have to rebase this. I missed the part about the order of the PRs.

arichiardi:
No worries, doing it now

Issue #134, opened 2018-02-21
Dumb terminal warning


javierjmx:
## Expected behavior
See the REPL prompt:
```
Clojure 1.9.0
                                                                                
user=> 
```

## Actual behavior
Seeing warning message and two prompts:
```

clojure: warning: Your terminal 'dumb' is not fully functional, expect some problems.

warnings can be silenced by the --no-warnings (-n) option
                                                                                
Clojure 1.9.0
                                                                                
user=> 
user=>
```

## Steps to reproduce the problem
Running `M-x inf-clojure` in a `.deps` project (clj).

### inf-clojure version information

2.1.0

### Emacs version

25.3

### Operating system

OSX 10.13.3

arichiardi:
Weird, the two prompts might be an actual parsing problem of the repl response.

PR #144, opened 2018-03-04
set inf-clojure-tools-deps-cmd to "clojure"

DogLooksGood:
Fix https://github.com/clojure-emacs/inf-clojure/issues/134


bbatsov:
Yep, `clj` should not be used with editors because of rlwrap.

PR #146, opened 2018-03-14
[security] Don't execute arbitrary functions

thblt:
Marking all values which validate `functionp` as "safe" means that any file can potentially execute any arbitrary code on the Emacs instance it's been opened on.  This commit should remove this possibility.

PS: I don't use this package myself: I'm trying to semi-automatically detect all occurrences of this issue (`:safe 'functionp`) on all Melpa packages, so apologies for any mistakes!  

Thanks!   

- [ ] (**Irrelevant**) You've updated the changelog (if adding/changing user-visible functionality)
- [ ]  (**Irrelevant**) You've updated the readme (if adding/changing user-visible functionality)

Thanks!

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -1417,7 +1417,6 @@ which is able to parse results in list form only.  You can peek
 at its implementation for getting to know some utility functions
 you might want to use in your customization."
   :type 'function
-  :safe #'functionp
   :package-version '(inf-clojure . "2.1.0"))
 
 (defconst inf-clojure-clojure-expr-break-chars "^[] \"'`><,;|&{()[@\\^]"


arichiardi:
This is great, however our packages relies on this behavior for. It will rely on it even more in the future. I see that your PR is a good thing though, @bohzidar?

thblt:
@arichiardi which package are you referring to?

arichiardi:
Sorry I meant this one.

thblt:
@arichiardi  I'm afraid there's a confusion there, so if you don't mind, I'm going to explain my change in more details.  Please don't take it bad if I'm stating the obvious, I just want to be sure we perfectly understand each other.  If this doesn't clarify the PR, could you explain me more precisely how the package relies on random files being able to execute random code without a warning?  I really don't understand this.  

 * `inf-clojure-completions-fn` is a Customize variable whose *type* is `functionp`.  This is how it works, and this PR *doesn't* change that.  Even after merging this PR, it will be perfectly possible to `(setq inf-clojure-completions-fn 'some-function`)

 * Emacs supports directory-local and file-local variables.  These variables are set automatically when a file with special markup is opened, *any* file.  A common use of these is to declare things like indentation length at a file or project level.   They look like [this](https://github.com/clojure-emacs/inf-clojure/blob/master/inf-clojure.el#L1610-L1613).

 * An obvious issue of this feature is security: most variables should only be set by the user, and **not** be available for any random file to modify.   

 * To solve this issue, `(defcustom)` has a `:safe` argument, which takes a predicate.  If the value of a dir-local/file-local variable validates this predicate, then the variable gets set **silently**.  `indent-tabs-mode`, for example, is perfectly safe to set to a boolean value in any file!  If the value *doesn't* validate the predicate (ie, the predicate returns nil), the user is warned that this value *may* be unsafe and asked if they want to set it, or to cancel.  

 * The current state of inf-clojure means that any file, simply by being opened, could:

     - assign a value to `inf-clojure-completions-fn`.  Any value which is a function will be silently accepted by Emacs.
     - wait until the mode executes the function pointed to by `inf-clojure-completions-fn`
     - profit, however that is: steal credentials, mine bitcoins, whatever :-)

 * My change means there's an attempt to set `inf-clojure-completions-fn` through dir-locals or file-local variables, the user will receive a warning that something fishy may be happening and will be asked to confirm that's actually what they want.  This leaves open the possibility of locally setting this variable without permitting to silently execute arbitrary code.

arichiardi:
Yes so I understand this and the concern very well, and I am torn because in my view `inf-clojure` should be completely customizable with elisp functions...it is a feature, not a bug.

I have planned a whole kind of framework that works off of customizing by calling your parsing function(s), because at the end of the day that is what the package does: provide hooks to repl evaluation.

Having said that, maybe there is another mechanism that I am not aware of for specifying behaviors. I saw that not even `unsafep` is enough for this in projectile. Indeed asking confirmation every time seems poor user experience...I am still unsure on what to do. I feel you are right, I wish we had an alternative.

thblt:
> Indeed asking confirmation every time seems poor user experience...

This should not be of concern, for two reasons:

 1. Emacs provide a mechanism to override these warnings:

    ``` elisp
    (setq enable-local-variables :all)
    ```

 2. It seems reasonable that these settings be configured in *directory*, not file, locals (ie in `.dir-locals.el`).  In this case, the confirmation prompt will happen exactly once per Emacs session.

-----

To be completely honest, I'm dissatisfied myself with the way file/dir-local variables security is implemented, and I think a solution should be sought.  My idea would be either:

 - to allow users to gpg-sign their `.dir-locals.el`.  If signature matches a list of pre-accepted PGP keys, all variables would be applied silently.   I don't have time to work on this right now, but it should be easy enough to implement.  

 - To give the option to consider this file safe across session.  In this case, just store an alist of (file-path . file-hash) somewhere and verify each `.dir-locals.el` before warning.  

This would allow to prevent unwanted warning without abusing the semantics of `:safe`.

arichiardi:
Oh the gpg is a good one, isn't there a `gpg` mechanism already baked in Emacs? About this PR, I think it should go in, /cc @bbatsov ?

thblt:
>  isn't there a gpg mechanism already baked in Emacs? 

Kindof, but I've had to reimplement bits for verifying package signatures for [borg-queen](https://github.com/thblt/borg-queen).  It's really easy anyway, although it requires some precaution and a good test suite to avoid introducing false  security :-)

Also, don't underestimate the second solution, IMHO.   Signing is a bit overkill when you can just say "as long as this file doesn't change don't talk to me about it"

arichiardi:
Very true the second works too.

thblt:
@arichiardi  Thanks!

PR #147, opened 2018-03-21
Update instructions to setup lumo repl

priyatam:
thanks @arichiardi!

-----------------

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Clojure(Script) development:
 * Support for Lumo
 * Support for Planck
 
-For a more powerful/full-featured solution see [CIDER][].
+For a more powerful/full-featured solution see [CIDER](https://github.com/clojure-emacs/cider).
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -62,22 +62,26 @@ Add the following to your Emacs config to enable
...
+**Don't enable `inf-clojure-minor-mode` and `cider-mode` at the same time. They
+have overlapping functionality and keybindings and the result will be nothing
+short of havoc.**
 
 ## Usage
 
 Just invoke `M-x inf-clojure` or press `C-c C-z` within a Clojure source file.
 This will start a REPL process for the current project and you can start
-interacting with it.
+interacting with it. By defaul this will look for `lein` command on the path.
```
arichiardi: This is not accurate. The logic is here, all the projects are tried in order, defaulting to generic: https://github.com/clojure-emacs/inf-clojure/blob/master/inf-clojure.el#L589

README.md
```diff
@@ -167,6 +171,32 @@ or the following to your [Emacs init file][]:
 
 The socket server REPL configuration options are described [here](https://dev.clojure.org/display/design/Socket+Server+REPL).
 
+#### Lumo Socket REPL
```
arichiardi: Can this section follow the same format as the other three? For coherence I guess. We also don't need to tell here about the devtools I guess. People should be already aware of that..

README.md
```diff
@@ -301,7 +331,7 @@ For example, you can use the following command (assuming `cp` contains
 the classpath) in your `.dir-locals.el`:
 
 ```el
-((nil . (eval . (setq inf-clojure-boot-cmd (concat "lumo -d -c "
+((nil . (eval . (setq inf-clojure-generic-cmd (concat "lumo -d -c "
```
arichiardi: The alignment of the second line is broken now, you probably need to change both

PR #157, opened 2019-01-19
[Fix #57] Add `inf-clojure-meta`

### LINKED ISSUE #57: Add an interactive command to display a var's metadata
Metadata is something we need to inspect fairly often, so it'd be easy if there was some command do display this. Implementing this should be trivial.

charignon:
**Summary**

This commit addresses
https://github.com/clojure-emacs/inf-clojure/issues/57 (Add an interactive
command to display a var's metadata) by adding an interactive command `inf-clojure-meta` to show metadata of a form at point.

This command works with clojure, lumo and planck alike.

**Test Plan**

Manually tested the command through inf-clojure using clojure, lumo and planck. I put my cursor on a
form and ran it interactively, it displayed the metadata in the inf buffer.
Specifically I tested with `(with-meta [1 2 3] {:data "hello"})` which printed `{:data "hello"}` in the inf-clojure buffer.

-----------------
- [NO] You've updated the readme (if adding/changing user-visible functionality)
  - It seems that we generally document higher level features and this does not warrant a change to the readme, let me know if you think that's necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## master (unreleased)
 
 * [#135](https://github.com/clojure-emacs/inf-clojure/pull/135): Improve command sanitation code.
+* [#157](https://github.com/clojure-emacs/inf-clojure/pull/157): Add `inf-clojure-meta`
 
 ## 2.1.0 (2018-01-02)
 
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -1099,6 +1099,39 @@ If you are using REPL types, it will pickup the most approapriate
 
 (define-obsolete-variable-alias 'inf-clojure-macroexpand-1-command 'inf-clojure-macroexpand-1-form "2.0.0")
 
+;;; Metadata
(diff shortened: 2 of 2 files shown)


charignon:
Could you suggest an example of syntax quirk that you think would be problematic?

arichiardi:
I might need to think about this through and see what other functions are doing, sorry, maybe it is actually already alright

arichiardi:
@charignon sorry for the delay, let me try this PR.

...
:computer: 
...

So I tried the branch, but I fail to see how it is going to work at the moment. I was expecting so do something like:

```clojure
(require '[clojure.string :as string])

cljs.user=> |clojure.string/blank?
```

and see the meta but that did not happen because we are missing the `var` around the `meta` call like ` (meta (var clojure.string/blank?))`

I feel I am misunderstanding the use case you want to address here (in which case, my fault :smile:)

bbatsov:
This hasn't had any activity in years, so I'll close it.

PR #164, opened 2019-05-31
Fix for eldoc-mode for ClojureCLR

dotemacs:
When using the REPL, whenever you'd type the function name, because
`Throwable` is JVM specific, `eldoc-mode` would try to look up the
function and when it wouldn't find it, it would bomb out with the
complaint that `Throwable` wasn't defined. Which is to be expected since
it's not running on JVM.

Looking at the documentation
https://docs.oracle.com/javase/10/docs/api/java/lang/Throwable.html

> The Throwable class is the superclass of all errors and exceptions in the Java language.

And in C#, Exception
https://docs.microsoft.com/en-us/dotnet/api/system.exception?redirectedfrom=MSDN&view=netframework-4.8

> This class is the base class for all exceptions.

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## master (unreleased)
 
+# [#164](https://github.com/clojure-emacs/inf-clojure/pull/164): Fix for eldoc-mode on ClojureCLR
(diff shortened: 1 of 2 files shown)


bbatsov:
Nice! Thanks for fixing this. I never thought that inf-clojure might be a good option for ClojureCLR users.

PR #165, opened 2019-06-02
[Fix #152]  Sanitize should only remove whitespace at the end of a command

### LINKED ISSUE #152: inf-clojure--make-single-line removes consecutive spaces in strings.
https://github.com/clojure-emacs/inf-clojure/blob/master/inf-clojure.el#L321

```(count "1   5") => 3 ;; Should be 5.```

EDIT: There are 3 spaces between 1 and 5 above. Apparently, this markdown is also removing the consecutive spaces.

### PR DESCRIPTION
**Pull request by arichiardi**
This patch makes sure that we trim at the right of the sanitized command only.

I am open to get rid completely of the sanitation but I do not have the time
now for trying on all the platform.

-----------------

- [] You've updated the changelog (if adding/changing user-visible functionality)
- [] You've updated the readme (if adding/changing user-visible functionality)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,6 @@ default: &default-steps
   steps:
     - checkout
     - run: apt-get update && apt-get install make
-    - run: make elpa
     - run: make test
 
 # Enumerated list of Emacs versions
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,19 @@
+.PHONY: clean version test
+
+all: build
+
+clean:
+	rm -rf .cask
+
+.cask:
+	cask install
+	cask update
+
 version:
 	emacs --version
 
-test : version
-	cask exec buttercup -L .
+build: version .cask
+	cask build
 
-elpa:
-	cask install
-	cask update
+test: version .cask
+	cask exec buttercup -L .
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -319,27 +319,15 @@ It requires a REPL PROC for inspecting the correct type."
     inf-clojure-repl-type))
 
 (defun inf-clojure--whole-comment-line-p (string)
-  "Return true iff STRING is a whole line semicolon comment."
+  "Return non-nil iff STRING is a whole line semicolon comment."
   (string-match-p "^\s*;" string))
-
-(defun inf-clojure--make-single-line (string)
-  "Convert a multi-line STRING in a single-line STRING.
-It also reduces redundant whitespace for readability and removes
-comments."
-  (let* ((lines (seq-filter (lambda (s) (not (inf-clojure--whole-comment-line-p s)))
-                            (split-string string "[\r\n]" t))))
-    (mapconcat (lambda (s)
-                 (if (not (string-match-p ";" s))
-                     (replace-regexp-in-string "\s+" " " s)
-                   (concat s "\n")))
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

**Comment by dotemacs** [comment]
> I do not have the time now for trying on all the platform.

In order to help, I've tested:

```clojure
(count "1   5")
```
on

- ClojureCLR 1.9.0
- Clojure 1.7.0
- Clojure 1.8.0
- Clojure 1.9.0
- Clojure 1.10.0
- Lumo 1.7.0
- Planck 2.10.0

They all return **5** as you'd expect.

bbatsov:
And multi-line evaluations don't mess up the REPL prompt? I recall this conversation was added to avoid having to deal with subprompts that appeared when entering multi-line expressions.

arichiardi:
Thanks for testing! There is also a test 😀

dotemacs:
> And multi-line evaluations don't mess up the REPL prompt? I recall this conversation was added to avoid having to deal with subprompts that appeared when entering multi-line expressions.

This is how it looks for Clojure 1.(7|8|9|10).0 & ClojureCLR 1.9.0:

```
eg.core=> (println 
"foo")
     #_=> foo
nil
eg.core=> 
```

Where I enter <kbd>C-q C-j</kbd> after `println`, followed by RET at the end of the s-expression.

Tried it out with Lumo & Planck and they behaved OK, but only once(!) the `inf-clojure-repl-type` was specified prior to the invocation of their REPLs, via .dir-locals.el:

```emacs-lisp
((nil . ((inf-clojure-lein-cmd . "lumo -r")
         (inf-clojure-minor-mode . t)
         (inf-clojure-repl-type . lumo))))
```

For me, I only need to use this mode for `ClojureCLR`. For the Clojure on JVM I use CIDER. 
I don't really use Planck & I've got some scripts in Lumo, but don't do it on daily basis.
I'd much rather that somebody who is "knee deep" in them check how their REPLs behave for them.

For Joker, I don't use it at all. Even with the above .dir-locals.el modified for Joker's settings, it was still trying to invoke JVM related documentation and throw errors. But once I disabled `eldoc-mode`, the REPL behaved just fine. Still, I'd like somebody who is into Joker with this mode to bless this before you merge the change.

arichiardi:
I am trying that `(count...)` thing working with `lumo` but this is what I see with this patch:

```

cljs.user=> ("count" "counted?")
cljs.user=> cljs.user=>        #_=>          #_=>                                        #_=>   ("count" "counted?")
cljs.user=> cljs.user=>          #_=>                                        #_=>   cljs.user=>        #_=>          #_=>                                        #_=>   ("test")
cljs.user=> (count "test  ")
6
cljs.user=>          #_=>                                        #_=>   ("count" "counted?")
cljs.user=> cljs.user=> (count "11111")
```

It seems like each and every `\n` is generating a subprompt. Probably it is a side-effect and can be treated in some of the post-filter functions that `comint` provides. It definitely rings a bell. I saw this problem already.

bbatsov:
Yeah, that's exactly the problem I tried to solve originally by just converting every expression to a single-line form. Probably some post-filter is going to be a better approach, though. I guess it's a common problem with most REPLs/shells. The visual artefacts weren't the real problem - this actually messes up tooling evals as well.

arichiardi:
It does, and in `lumo`'s case it actually crashes the REPL.

PR #170, opened 2020-04-12
Insert commands

dpsutton:
Nice interaction to see the form and then the output.

I'm a big fan of the insert in the repl commands so adding them here.

 - 14 errors before me. Will clean up in another pass
  - doesn't seem like an obvious place to put this right now

Thanks!


bbatsov:
Nice work! Thanks!

PR #172, opened 2020-04-15
Dispatch

dpsutton:
Here's a preview of what I think inf-clojure could/should look like right now. I wanted to address a few issues:

1. when i started up a planck repl, it was constantly telling me that get-completions is private
2. using a cljs repl was also trying to get completions and dying. Random features didn't work. 

Looking through the code, I didn't have an easy way to see which repl types supported which features and what to override to make it work. Also, this dispatch system is fundamentally closed with the `pcase` statements rather than being open. This means no one can easily add a specific `bb` repl and fundamentally `clojure -m cljs.main -r` wasn't supported out of the box.

I think inf-clojure should offer a standard interface and never throw an error. It should be incredibly clear which features a repl supports and how to override them. It should be trivial to extend and support more repl types.

For sure I probably have some styling and the README needs a lot of love. I also want to do some screencasts showing how to use inf-clojure. Its incredibly easy and nice and I want others to see how nice it is.

An example of why this appeals to me is `inf-clojure-cljs.el` which defines a mode for regular `clojure -m cljs.main -r`:

```emacs-lisp
(defun inf-clojure-cljs-features-dispatch (feature)
  (case feature
    ;; ((load) "(cljs.core/load-file \"%s\")")
    ((doc) "(cljs.repl/doc %s)")
    ((source) "(cljs.repl/source %s)")
    ((arglists) "(try (->> '%s cljs.core/resolve cljs.core/meta :arglists) (catch :default _ nil))")
    ((apropos) "(cljs.repl/apropos \"%s\")")
    ((ns-vars) "(cljs.repl/dir %s)")
    ((set-ns) "(in-ns '%s)")
    ((macroexpand) "(cljs.core/macroexpand '%s)")
    ((macroexpand-1) "(cljs.core/macroexpand-1 '%s)")
    ;; ((completion) inf-clojure-completion-form-lumo)
    ))

 'cljs  #'inf-clojure-cljs-features-dispatch "cljs.repl" "clojure -m cljs.main -r")
```

This makes it clear which features are supported and which aren't, how to change them, and a default command to get this type of repl up and running.

For this first pass these aren't overrideable but these just need to be extracted to defcustoms. But this form is what I wanted to write from the beginning.

I also am not super happy with this form of polymorphism but I don't know a better way in elisp. I want a protocol but don't know how to do it here. This seems fine for now i guess.
-----------------

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ You can install `inf-clojure` using the following command:
 
 or if you'd rather keep it in your dotfiles:
 
-```el
+```emacs-lisp
 (unless (package-installed-p 'inf-clojure)
   (package-refresh-contents)
   (package-install 'inf-clojure))
@@ -60,7 +60,7 @@ If the installation doesn't work try refreshing the package list:
 Add the following to your Emacs config to enable
 `inf-clojure-minor-mode` for Clojure source buffers:
 
-```el
+```emacs-lisp
 (add-hook 'clojure-mode-hook #'inf-clojure-minor-mode)
 ```
 
@@ -70,39 +70,27 @@ short of havoc.**
 
 ## Usage
 
-Just invoke `M-x inf-clojure` or press `C-c C-z` within a Clojure source file.
-This will start a REPL process for the current project and you can start
-interacting with it.
+Just invoke `M-x inf-clojure` or press `C-c C-z` within a Clojure
+source file. You should get a prompt with the supported repl types and
+common startup forms. You can select one of these or type in your own
+custom startup. This will start a REPL process for the current project
+and you can start interacting with it.
 
-Inf-clojure has several custom variables which control the command used to
-start a REPL for a particular project type:
-
- - `inf-clojure-lein-cmd` ([Leiningen][])
- - `inf-clojure-boot-cmd` ([Boot][])
- - `inf-clojure-tools-deps-cmd` ([Clojure cli tools][])
- - `inf-clojure-generic-cmd`
-
-Detection is attempted in the above order (see [function
-`inf-clojure-project-type` in this
-file](https://github.com/clojure-emacs/inf-clojure/blob/master/inf-clojure.el#L611-L618)
-but the `inf-clojure-project-type` variable can force a particular
-project type, useful for projects that don't have standard layouts.
+If you've already started a socket repl, use `M-x inf-clojure-connect`
+and enter the host and port numbers.
 
 It is highly recommended to use a cons pair like `("localhost" . 5555)` to
 connect to a socket REPL, terminal REPLs are inherently hard to work with and
 support will be deprecated in the foreseeable future.
 
-Interactively, use `M-x inf-clojure-connect` (`C-c M-c`) to connect to a
-running socket REPL or `C-u C-c C-z` for specifying a different command/cons
-pair.
+Inf-clojure aims to be very simple and offer tooling that the repl
+itself exposes. A few commands are:
 
-You can also set custom values to `inf-clojure` variables on a per-project
-basis using
-[directory variables](https://www.gnu.org/software/emacs/manual/html_node/emacs/Directory-Variables.html).
-
-For a list of all available commands in `inf-clojure-mode` (a.k.a. the REPL) and
-`inf-clojure-minor-mode` you can either invoke `C-h f RET inf-clojure-mode` and
-`C-h f RET inf-clojure-minor-mode` or simply browse their menus.
+- eval last sexp `C-x C-e`
+- show arglists for function `C-c C-a`
+- show var documentation `C-c C-v`
+- show source `C-c C-s`
+- insert top level form into repl `C-c C-j d`
 
 Many `inf-clojure-minor-mode` commands by default act on the symbol at
 point. You can, however, change this behaviour by invoking such
@@ -121,13 +109,13 @@ clojure -J-Dclojure.server.repl="{:port 5555 :accept clojure.core.server/repl}"
 
 Then either `C-c M-c RET localhost RET 5555` from within Emacs or add the following to your `.dir-locals.el`:
 
-```el
+```emacs-lisp
 ((nil . ((inf-clojure-tools-deps-cmd . ("localhost" . 5555)))))
 ```
 
 or the following to your [Emacs init file][]:
 
-```el
+```emacs-lisp
 (setf inf-clojure-tools-deps-cmd '("localhost" . 5555)):
 ```
 
@@ -139,38 +127,14 @@ For Leiningen, add the following option to your `~/.lein/profiles.clj` or your `
 :jvm-opts ["-Dclojure.server.repl={:port 5555 :accept clojure.core.server/repl}"]
 ```
 
-Then run `lein repl` from within your project directory to start the REPL, and either `C-c M-c RET localhost RET 5555` from within Emacs or add the following to your `.dir-locals.el`:
+Then run `lein repl` from within your project directory to start the
+REPL.  To connect, you can either `m-x inf-clojure-connect [RET]
+localhost [RET] 5555` or you can put in a dir local file the
+information on how connect:
 
-```el
-((nil . ((inf-clojure-lein-cmd . ("localhost" . 5555)))))
+```emacs-lisp
+((nil (inf-clojure-custom-startup "localhost" . 5555)))
 ```
-
-or the following to your [Emacs init file][]:
-
-```el
-(setf inf-clojure-lein-cmd '("localhost" . 5555))
-```
-
-#### Boot Socket REPL
-
-For boot, export the environment variable `BOOT_JVM_OPTIONS`:
-
-```shell
-export BOOT_JVM_OPTIONS='-Dclojure.server.repl="{:port 5555 :accept clojure.core.server/repl}"'
-```
-
-or add the following to your `.dir-locals.el`:
-
-```el
-((nil . ((inf-clojure-boot-cmd . ("localhost" . 5555)))))
-```
-
-or the following to your [Emacs init file][]:
-
-```el
-(setf inf-clojure-boot-cmd '("localhost" . 5555))
-```
-
 The socket server REPL configuration options are described [here](https://clojure.org/reference/repl_and_main#_launching_a_socket_server).
 
 #### Lumo Socket REPL
@@ -181,19 +145,19 @@ For example if a `project.clj` is present in the project root folder, `inf-cloju
 
 After you launch `lumo ... -n 5555`, as customary, either `C-c M-c RET localhost RET 5555` from within Emacs or add the following to your `.dir-locals.el`:
 
-```el
+```emacs-lisp
 ((nil . ((inf-clojure-lein-cmd . ("localhost" . 5555)))))
 ```
 
 or the following to your [Emacs init file][]:
 
-```el
+```emacs-lisp
 (setf inf-clojure-lein-cmd '("localhost" . 5555))
 ```
 
 Project detection can be completely skipped and the `generic` project type can be used instead:
 
-```el
+```emacs-lisp
 (setf inf-clojure-project-type . "generic")
 (setf inf-clojure-generic-cmd '("localhost" 5555))
 ```
@@ -206,7 +170,7 @@ filter out ASCII escape characters at the moment and will not behave correctly.
 
 You can disable coloring the following way for `boot`:
 
-```el
+```emacs-lisp
 ((nil . ((inf-clojure-boot-cmd . "boot repl -C"))))
 ```
 
@@ -286,7 +250,7 @@ By default `inf-clojure` would start a standard Clojure REPL using
 right launch command (or connect to the REPL via a socket).  For example, for
 Lumo just add the following in your `.dir-locals.el`:
 
-```el
+```emacs-lisp
 ((nil . ((inf-clojure-boot-cmd . "lumo -d")))) ;; inf-clojure-lein-cmd if you are using Leiningen
 ```
 
@@ -302,7 +266,7 @@ echo area.
 You can activate ElDoc with `M-x eldoc-mode` or by adding the
 following to you Emacs config:
 
-```el
+```emacs-lisp
 (add-hook 'clojure-mode-hook #'eldoc-mode)
 (add-hook 'inf-clojure-mode-hook #'eldoc-mode)
 ```
@@ -332,12 +296,25 @@ in order to play nicely with emacs.
 For example, you can use the following command (assuming `cp` contains
 the classpath) in your `.dir-locals.el`:
 
-```el
+```emacs-lisp
 ((nil . (eval . (setq inf-clojure-generic-cmd (concat "lumo -d -c "
                                                       (f-read (concat (inf-clojure-project-root) "cp")))))))
 ```
 
 ## Troubleshooting
+
+### Things seem broken
+
+Inf-clojure is intentionally quite simple and just sends commands to a
+repl on your behalf to provide features. In order to do this
+inf-clojure largely needs to know the repl type so it can format the
+correct calls. Most end up in `(lumo.repl/doc [symbol])` or
+`(cljs.repl/doc ...)` so its important that the repl type is set
+correctly. This repl type exists in the process buffer (repl) and the
+source buffers as a cache. If you have problems, run `m-x
+inf-clojure-set-repl-type` from the source buffer to set the repl type
+in both buffers. To see how simple inf-clojure is, look at
+`inf-clojure-repl-features` to see largely how things are laid out.
 
 ### REPL not responsive in Windows OS
 
@@ -358,7 +335,7 @@ jline.terminal=unsupported
 
 Standard Emacs debugging turns out to be difficult when an asynchronous process is involved. In this case try to enable logging:
 
-```el
+```emacs-lisp
 (setq inf-clojure-log-activity t)
 ```
 
--- a/README.md
+++ b/README.md
@@ -228,22 +228,46 @@ configurable.
 You can see all the configuration options available using the command
 `M-x customize-group RET inf-clojure`.
 
+The supported repl-features are in an alist called
+`inc-clojure-repl-features` and it has the following shape:
+
+```emacs-lisp
+'((cljs . ((doc . "(cljs.repl/doc %s)")
+           (source . "(cljs.repl/source %s)")
+           (arglists . "(try (->> '%s cljs.core/resolve cljs.core/meta :arglists) (catch :default _ nil))")
+           (apropos . "(cljs.repl/apropos \"%s\")")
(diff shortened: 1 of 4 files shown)


arichiardi:
Oh this is so so nice, it reminds a bit of lsp 😉

bbatsov:
I think splitting out the expressions for each backend in a separate source file is an overkill and would impact negatively the simplicity of `inf-clojure`. We don't really gain much by this, as we could easily group those better in the existing source file. 

I also wanted to get rid of most of the defcustoms, as when you come to think about it they serve almost no purpose - originally this mode supported only Clojure and it was kind of OK to have 10 defcustom around, even if most people probably never changed those. I think the entire config can be a single map (or rather some alist) similar to this:

```
;; ((load) "(cljs.core/load-file \"%s\")")
    ((doc) "(cljs.repl/doc %s)")
    ((source) "(cljs.repl/source %s)")
    ((arglists) "(try (->> '%s cljs.core/resolve cljs.core/meta :arglists) (catch :default _ nil))")
    ((apropos) "(cljs.repl/apropos \"%s\")")
    ((ns-vars) "(cljs.repl/dir %s)")
    ((set-ns) "(in-ns '%s)")
    ((macroexpand) "(cljs.core/macroexpand '%s)")
    ((macroexpand-1) "(cljs.core/macroexpand-1 '%s)")
    ;; ((completion) inf-clojure-completion-form-lumo)
    ))
```

With some top-level key for the implementation. I doubt many people would need to modify it, but they would certainly be able to do so. Which is also related to what you said:

> For this first pass these aren't overrideable but these just need to be extracted to defcustoms. But this form is what I wanted to write from the beginning.

> For sure I probably have some styling and the README needs a lot of love. I also want to do some screencasts showing how to use inf-clojure. Its incredibly easy and nice and I want others to see how nice it is.

Yeah, that'd be nice for sure. 

Btw, I love the list of ideas you've compiled. The only part the puzzled me was the one about the project detection and geiser is doing, as I haven't used it for a while and I didn't remember. CIDER's detection of multiple build tools and asking which to use when starting something seems like a pretty reasonable approach to me, as you don't want to have commands like `start-clojure-repl-with-boot`. Of course, for the sake of simplicity we can just drop this altogether and have people be prompted explicitly for some REPL startup command. For a simple mode we need a simple approach. :-)

bbatsov:
One more thing - I really think we should remove the REPL type detection and just have the user specify explicitly what type of REPL are they starting/connecting to. This is going to simplify things a lot IMO.

dpsutton:
First, thanks for all of your work and thoughts. This is largely a shifting of code and simplifying a ton of previous work and certainly is not anything novel.

> I think splitting out the expressions for each backend in a separate source file is an overkill and would impact negatively the simplicity of inf-clojure.

This is a great suggestion right to the core. I'm going to kill all of these defcustoms and have literal structures holding all of the interaction.

> I also wanted to get rid of most of the defcustoms, as when you come to think about it they serve almost no purpose

Gone and combined with the above.

> With some top-level key for the implementation. I doubt many people would need to modify it, but they would certainly be able to do so.

I'm kinda torn on this one. I'm not sure I see a reason people would want to override these unless we got it wrong. And it should be trivial to add new implementations I wonder if this is neccessary. For instance, someone could just overwrite the clojure versions with:

```emacs-lisp
(inf-clojure-register-implementation
 'clojure  #'my-own-values "clojure.core.server" "clojure")
```

Having that hashmap makes things easily overrideable and quick to lookup.

If not, I want to make sure things are overrideable from dir-locals. I think if all of this was just nested alists then that would work? Assuming a dir-locals file can bind `inf-clojure-implementations-alist` as some new value for lumo at the front of the existing list this should be fine.

> The only part the puzzled me was the one about the project detection and geiser is doing, as I haven't used it for a while and I didn't remember.

Geiser doesn't do any project detection and I'm not suggesting that either. Geiser instead installs the commands `run-guile`, `run-chicken`, etc. for all of the different schemes. I was thinking something similar for inf-clojure like `inf-run-lumo`, `inf-run-planck`, etc. And it would be cool if a user registered a new implementation for `bb` it would automatically intern `inf-run-bb`. Then when editing a babashka file, `m-x inf-run-bb` and they are up and running.

> Of course, for the sake of simplicity we can just drop this [project detection] altogether

I think I like this a lot. Especially since things these days are often running lein plugins or more commonly clj with different aliases involved. I think putting the user in touch with the startup command makes a lot of sense and it takes an incredibly amount of work to ensure this feature is not in the way, correct, and overrideable.

> One more thing - I really think we should remove the REPL type detection and just have the user specify explicitly what type of REPL are they starting/connecting to. This is going to simplify things a lot IMO.

I think the code that figures out which repl it is works pretty decently. I was thinking that we would make it much more clear which type we had inferred and i already made `m-x inf-clojure-set-repl-type` which does a completing read over all known interpretations. This way, if we get it wrong, its first and foremost clear what we have wrong, and easy to fix by an `m-x` command.

inf-clojure.el
```diff
@@ -343,6 +343,12 @@ number (e.g. (\"localhost\" . 5555))."
    (stringp (car x))
    (numberp (cdr x))))
 
+(defcustom inf-clojure-custom-startup
+   nil
+   "Form to be used to start inf-clojure.
```
bbatsov, line 348: That's a command, not a form, so I'd reflect it in the name of the defcustom.

inf-clojure.el
```diff
@@ -642,10 +648,11 @@ it (default is value of `inf-clojure-*-cmd').  Runs the hooks
 from `inf-clojure-mode-hook' (after the `comint-mode-hook' is
 run).
 \(Type \\[describe-mode] in the process buffer for a list of commands.)"
-  (interactive (list (completing-read "Clojure startup command: "
-                                      (mapcar #'cdr inf-clojure-startup-forms)
-                                      nil
-                                      'confirm-after-completion)))
+  (interactive (list (or inf-clojure-custom-startup
+                         (completing-read "Clojure startup command: "
```
bbatsov, line 652: I guess this is more of a "REPL startup command", although in the case of connecting it's different, so I feel the message is a bit confusing.

inf-clojure.el
```diff
@@ -700,29 +661,52 @@ Fallback to `default-directory.' if not within a project."
...
+             ;; that already has the window in it.
+             (or pop-up-frames
+                 (get-buffer-window inf-clojure-buffer t))))
+        (pop-to-buffer inf-clojure-buffer))
+    (call-interactively #'inf-clojure))
+  (when eob-p
+    (push-mark)
+    (goto-char (point-max))))
+
+(defun inf-clojure-quit (&optional buffer)
```
bbatsov, line 617: Seems to me command like this one should also check if it was invoked in a REPL buffer and kill it directly.

todo.org
```diff
@@ -50,9 +50,15 @@ The source primitive is quite nice but we most likely need a way to navigate to
...
-** TODO Better handling of cyclic dependencies
-In inf-clojure.el I load the implementation specific files for lumo, planck, etc _after_ defining the register function. I would much prefer a way to do this without being so fragile.
-
 ** TODO Put repl type in modeline
 Rather than just ~*inf-clojure*~ we could put the repl type. Make it easy to follow and makes it easy to see when it gets it wrong.
+
+** TODO How do CIDER and inf-clojure play nice on the same emacs?
+inf-clojure and CIDER are fighting over the keymappings. I've been doing a bit of a kludge to remove CIDER's tentacles from my clojure files for developing:
+#+BEGIN_SRC emacs-lisp
+  (seq-doseq (buffer (buffer-list))
```
bbatsov, line 62: Might be best to add check in the in the minor mode startup sequence which checks if cider-mode is present and disables it automatically.

todo.org
```diff
@@ -8,6 +8,9 @@ For some reason ~inf-clojure--set-repl-type~ is called in:
 
 Seems better to do this on the two different connection methods and then be done with it?
 
+** TODO do we need repl type in both source buffer and connection?
+   these can get out of sync and lead to confusing errors when closing a repl and opening a new one. It seems like we keep the repl-type in the source buffer to prevent a single ~(with-current-buffer (process-buffer proc) inf-clojure-repl-type)~
```
bbatsov, line 12: I think it's not needed in the source buffer. In the interest of simplicity I think we should go back to the concept of a "current-repl" that users select explicitly and afterwards all commands just interact with it. Then there's no need to track any relationships between source buffers and REPLs, although you can allow for `buffer-local` mappings to REPLs for conveniences sake. Static mapping is the best mapping! :D

inf-clojure.el
dpsutton: even if its a cons cell of host and port? Custom startup sounds vague enough that it encompasses both. If it had command I might think it was a string to start up rather than allowing for the connection tuple. I don't have strong opinions though so happy to take a final answer and commit it.

todo.org
dpsutton: Yeah i'm not sure what's going on with this one. I haven't touched the bottom half of the file. Not sure why its not using `(inf-clojure-proc)` which seems to be the standard way every where else

```emacs-lisp
  (let ((target-buffer (or buffer (inf-clojure-select-target-repl))))
    (when (get-buffer-process target-buffer)
      (delete-process target-buffer))
    (kill-buffer target-buffer))
```

todo.org
dpsutton: Yeah I've removed it.

README.md
```diff
@@ -47,7 +47,7 @@ You can install `inf-clojure` using the following command:
 
 or if you'd rather keep it in your dotfiles:
 
-```el
+```emacs-lisp
```
bbatsov, line 50: Are you sure about this? I've been using `el` for years and I think it's just an alias for `cl`. I guess `emacs-lisp` could be the same.

dpsutton: image
 Github seems to like it.

dpsutton: https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L1342

bbatsov: Good to know. Seems I've messed up something. :)

dpsutton:
What remains to be done? I'm noodling on how to allow the clojure completion form to be added when the completions library is on the classpath.

Blockers to release
1. Update readme, expunge all old stuff no longer relevant
2. finalize data structure for repl-features. I'm not sure the big nested alist is that great. There's almost no way to override anything if anyone wants. Not usually a big deal but for clojure, if someone includes complete.core on the classpath it would be nice to offer completions.

I'm gonna delete almost every test that doesn't give me any confidence. There aren't many to start with.

bbatsov:
> finalize data structure for repl-features. I'm not sure the big nested alist is that great. There's almost no way to override anything if anyone wants. Not usually a big deal but for clojure, if someone includes complete.core on the classpath it would be nice to offer completions.

I wouldn't fret much over this, as modifying alists in place is actually quite easy in Emacs - see https://emacs.stackexchange.com/questions/3397/how-to-replace-an-element-of-an-alist

We just a need to provide some API that reduces this to specifying repl type, feature and form, so people won't have to learn how to work with alists. :-) 

For the completion - it can simply be wrapped in some conditional require. Note that Lein at least will likely drop `clojure-complete` when nREPL 0.8 is out, as it features built-in completion, but for the time being it's a reasonable option to check for it. I'm think another simple option would be to ship with `inf-clojure` some clojure file that gets loaded on start, which provides things like simple completion and so on. That would reduce the need to inlined Clojure code and feature presence checks.

dpsutton:
@bbatsov I think i'm ready for this to merge. I've updated the completion part of the readme and added an API so the features can be mutated. Its not quite the same as its not friendly to per-project changes but I think i'm happy for the moment.

The sum is that the implementations are all in one place and easy to see. We've added clarity to the codebase. The startup is easier now and shows default commands. The implementations has been extended to cover cljs and bb in the meantime.

README.md
```diff
@@ -70,39 +82,27 @@ short of havoc.**
...
-You can also set custom values to `inf-clojure` variables on a per-project
-basis using
-[directory variables](https://www.gnu.org/software/emacs/manual/html_node/emacs/Directory-Variables.html).
+Inf-clojure aims to be very simple and offer tooling that the repl
+itself exposes. A few commands are:
 
-For a list of all available commands in `inf-clojure-mode` (a.k.a. the REPL) and
-`inf-clojure-minor-mode` you can either invoke `C-h f RET inf-clojure-mode` and
-`C-h f RET inf-clojure-minor-mode` or simply browse their menus.
+- eval last sexp `C-x C-e`
```
arichiardi, line 101: Better!

README.md
```diff
@@ -312,6 +271,18 @@ You can leave it enabled, it just won't show anything in the echo area.
 
 #### Code Completion
 
+This section is out of date. Code completion used to work only in lumo
```
arichiardi, line 274: This is a bit confusing for me, so if there is no completion for a repl I thought we would just not be completing?

inf-clojure.el
arichiardi: Maybe we can drop "Clojure" here?

inf-clojure.el
```diff
@@ -809,15 +829,16 @@ The prefix argument SWITCH-TO-REPL controls whether to switch to
...
 is present it will be used instead of the current file."
   (interactive "P")
-  (let ((proc (inf-clojure-proc))
-        (file-name (or file-name
-                       (car (comint-get-source "Load Clojure file: " inf-clojure-prev-l/c-dir/file
-                                               ;; nil because doesn't need an exact name
-                                               inf-clojure-source-modes nil)))))
+  (let* ((proc (inf-clojure-proc))
+         (file-name (or file-name
+                        (car (comint-get-source "Load Clojure file: " inf-clojure-prev-l/c-dir/file
```
arichiardi, line 834: I would also drop Clojure here, it is kind of clear from the package context ...one day we should actually customize it by repl type (like joker actually loads `.joker` files)

todo.org
```diff
@@ -0,0 +1,80 @@
...
+2. inf-clojure-reload-form
+3. inf-clojure-reload-all-form
+
+Seems better to do this on the two different connection methods and then be done with it?
+
+** DONE do we need repl type in both source buffer and connection?
+   these can get out of sync and lead to confusing errors when closing a repl and opening a new one. It seems like we keep the repl-type in the source buffer to prevent a single ~(with-current-buffer (process-buffer proc) inf-clojure-repl-type)~
+
+** TODO nice startup
+There's some project detection but that's becoming less and less useful as time goes on. Shadow, lein, deps.edn can all easily be mixed in the same project. And then lumo, planck, or bb scripts could live side by side. Rather than trying to guess the project type, I think i'd like to mimic geiser's style of handling multiple scheme backends. Perhaps ~m-x inf-clojure-run-planck~ and similar could help out.
```
arichiardi, line 15: I would say that project detection is very tricky cause for instance lumo files could still use a project.clj file for generating the classpath.

I would actually start considering static file/buffer repl detection (instead of find-ns we identify namespaces in the file content)

README.md
dpsutton: that's correct. the old way tried to complete if a random library was on the classpath and required. This was often the case with lein but i'm not sure that's a majority of startups these days. So its removed and this section explains how to get it back if desired.

todo.org
dpsutton: yeah these are just sketches for a nicer startup. and its for future work :)

README.md
arichiardi: Ok cool tnx I did not remember properly then 😁

README.md
bbatsov: Btw, I figure out why I was confused - turns out it's `el` in pygments (e.g. when you're using RTD or MkDocs). Too many doc systems. :D

todo.org
bbatsov: Well, I meant to change many things that I never got to changing, so this was probably one of them. :-)

bbatsov:
@dpsutton I'm fine with merging in the current form, although I'll also leave a couple of small remarks. 

Still, I'd suggest cleaning up the commit log, as I think you can reduce the important changes to something like 5-10 commits.

README.md
```diff
@@ -8,10 +8,22 @@
...
-extends `clojure-mode` with commands to evaluate forms directly in the
-REPL.
+`inf-clojure` aims to expose the extensive self-documenting features
+of Clojure repls into an emacs package. Inf-clojure does not require
+middleware or complicated machinery, but it more a predefined set of
+commands for a repl. Because of this, you can interact with many more
+repl types than an nREPL-based tool like CIDER. Inf-clojure supports
+the following repls:
+- Clojure
+- clojurescript through cljs.main (no piggieback required)
```
bbatsov, line 18: I don't think the clarification here is needed, as without nREPL obviously you don't need piggieback (at least it seems obvious to me). Might a good idea to capitalize all names consistently in this list.

README.md
```diff
@@ -8,10 +8,22 @@
...
+- planck
+- lumo
+- joker
+- bb
```
bbatsov, line 22: I'd spell out Babashka here, as `bb` probably means nothing to people who haven't used it.

README.md
```diff
@@ -70,39 +82,27 @@ short of havoc.**
...
-running socket REPL or `C-u C-c C-z` for specifying a different command/cons
-pair.
-
-You can also set custom values to `inf-clojure` variables on a per-project
-basis using
-[directory variables](https://www.gnu.org/software/emacs/manual/html_node/emacs/Directory-Variables.html).
+Inf-clojure aims to be very simple and offer tooling that the repl
+itself exposes. A few commands are:
 
-For a list of all available commands in `inf-clojure-mode` (a.k.a. the REPL) and
```
bbatsov, line 103: It seems to me this paragraph and the one about `.dir-locals.el` are still useful.

README.md
```diff
@@ -312,6 +271,18 @@ You can leave it enabled, it just won't show anything in the echo area.
...
+and in lein Clojure repls. It would throw errors almost everywhere
+else. To restore the old behavior you can do the following:
+
+```emacs-lisp
+(inf-clojure-update-feature 'clojure 'completion "(complete.core/completions \"%s\")")
+```
+
+This expects there to be a complete.core required and on the
+classpath. You could alternatively use `compliment {:mvn/version
+"0.3.10"}` and then use an alias to ensure this is on your classpath.
```
bbatsov, line 284: You might also add an example snippet about using `compliment`.

dpsutton:
Squashed, readme updated, and I think we are good to go.

bbatsov:
Indeed! Great work!

dpsutton:
Oh I should have updated `inf-clojure-version`. its currently `"2.2.0"`  where should the snapshot go? 2.3? 3.0 due to the defcustoms being removed?

dpsutton:
actually no need to respond. i see you've already taken care of that. thanks!

bbatsov:
Btw, at some point all the major changes should be listed in the changelog as well.

PR #174, opened 2020-04-25
Set REPL type from startup form, prompt, or defcustom at startup

dpsutton:
Set REPL type at startup from startup form. Defcustom to override or set if custom startup form is used and otherwise prompt user for repl type.

Thanks!

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -71,14 +71,6 @@
 (require 'subr-x)
 (require 'seq)
 
-(defvar inf-clojure-recognize-alist '((lumo . "lumo.repl")
-                                      (planck . "planck.repl")
-                                      ;; cljs goes after the selfhosts
-                                      (cljs . "cljs.repl")
-                                      (joker . "joker.repl")
-                                      (babashka . "babashka.classpath")
-                                      (clojure . "clojure.core.server")))
-
 (defvar inf-clojure-startup-forms '((lein . "lein repl")
                                     (boot . "boot repl")
                                     (clojure . "clojure")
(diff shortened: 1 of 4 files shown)


todo.org
```diff
@@ -81,3 +77,9 @@ Seems a bit heavy handed but its working for me so far.
 
 ** TODO is disabling color still required?
    in the readme it mentions that color should be turned off. in my usage I haven't run into this problem at all. perhaps no longer true?
+** TODO nice startup
+There's some project detection but that's becoming less and less useful as time goes on. Shadow, lein, deps.edn can all easily be mixed in the same project. And then lumo, planck, or bb scripts could live side by side. Rather than trying to guess the project type, I think i'd like to mimic geiser's style of handling multiple scheme backends. Perhaps ~m-x inf-clojure-run-planck~ and similar could help out.
+
+Some considerations:
+- is this path aware? IE, don't show an option to run planck, lumo, etc, if they aren't visible or installed?
```
arichiardi, line 84: Imho to keep it simple not to have path inspection is a good start.

todo.org
```diff
@@ -81,3 +77,9 @@ Seems a bit heavy handed but its working for me so far.
...
+- should it have a rebuild function so that user registered implementations can show up in the ~m-x~ menu as well?
```
arichiardi, line 85: If we go for the classic defcustom as list approach and maybe a register function like `lsp` imho yes

inf-clojure.el
```diff
@@ -217,37 +209,13 @@ either `setq-local` or an entry in `.dir-locals.el`." )
...
   (interactive)
-  (let* ((proc (inf-clojure-proc))
-         (types (mapcar #'car inf-clojure-repl-features))
-         (type-to-set (intern
-                       (completing-read "Set REPL type:"
-                                        (sort (mapcar #'symbol-name types) #'string-lessp)))))
-    (with-current-buffer (process-buffer proc)
-      (setq-local inf-clojure-repl-type  type-to-set))))
-
-(defun inf-clojure--set-repl-type (proc)
```
arichiardi, line 242: Farewell, it has served as well-ish 😁

PR #175, opened 2020-04-25
Mark this is not part of emacs. probably copy paste error

dpsutton:
looks like this got copy pasted from some emacs lisp files which were part of emacs


bbatsov:
Great catch! Yeah, initially I just copy-pasted `inferior-lisp` and started tweaking it.

PR #179, opened 2020-05-16
[Fix #178] Ensure there is a directory for comint (thanks mcphailm)

### LINKED ISSUE #178: (wrong-type-argument stringp nil)
## Expected behavior

working inf-clojure mode

## Actual behavior

getting `(wrong-type-argument stringp nil)` on load and `Error running timer 'auto-revert-buffers (wrong-type-argument string nil)` after that

1) load any .clj file
2) M-x inf-clojure
3) select any repl (tried with bb and clojure)

inf-clojure 20200425.1648

### Lein/Boot version

n/a

### Emacs version

emacs 26.3

### Operating system

arch linux

dpsutton:
If there was no containing directory with a project file the
default-directory would be nil. This leads to problems in starting up
inf-clojure and also annoyingly seems to set an error in a timer:

> Error running timer 'auto-revert-buffers (wrong-type-argument string nil)

Fixes https://github.com/clojure-emacs/inf-clojure/issues/178

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -667,7 +667,7 @@ process buffer for a list of commands.)"
                                           'confirm-after-completion))))
   (if (not (comint-check-proc "*inf-clojure*"))
       ;; run the new process in the project's root when in a project folder
-      (let ((default-directory (clojure-project-dir))
+      (let ((default-directory (or (clojure-project-dir) default-directory))
             (cmdlist (if (consp cmd)
                          (list cmd)
                        (split-string cmd)))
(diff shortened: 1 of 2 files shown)


inf-clojure.el
```diff
@@ -667,7 +667,7 @@ process buffer for a list of commands.)"
                                           'confirm-after-completion))))
   (if (not (comint-check-proc "*inf-clojure*"))
       ;; run the new process in the project's root when in a project folder
-      (let ((default-directory (clojure-project-dir))
+      (let ((default-directory (or (clojure-project-dir) "./"))
```
bbatsov, line 670: Why not just to default to `default-directory` in this case?

dpsutton: great point! thanks!

bbatsov: I see my remark as resolved, but the code seems the same to me.

dpsutton: weird. i thought i changed this and force pushed. i'll fix it tonight. sorry!

bbatsov: No worries!

PR #180, opened 2020-05-16
Ensure comint-truncate-buffer call is from the repl buffer

dpsutton:
When clearing comint buffer ensure we are actually in the repl buffer

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -598,8 +598,11 @@ to continue it."
 (defun inf-clojure-clear-repl-buffer ()
   "Clear the REPL buffer."
   (interactive)
-  (let ((comint-buffer-maximum-size 0))
-    (comint-truncate-buffer)))
+  (with-current-buffer (if (derived-mode-p 'inf-clojure-mode)
+                           (current-buffer)
+                         inf-clojure-buffer)
+    (let ((comint-buffer-maximum-size 0))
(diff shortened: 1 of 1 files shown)


inf-clojure.el
```diff
@@ -598,8 +598,11 @@ to continue it."
 (defun inf-clojure-clear-repl-buffer ()
   "Clear the REPL buffer."
   (interactive)
-  (let ((comint-buffer-maximum-size 0))
-    (comint-truncate-buffer)))
+  (with-current-buffer (if (derived-mode-p 'inf-clojure-mode)
```
bbatsov, line 601: You don't really need `with-current-buffer` here. I'm guessing we can also show some informative message if the user tries to invoke this in the wrong buffer, although that's not a big deal IMO.

dpsutton: I did this because i don't consider source buffers the wrong buffer. I wanted to be able to clear the repl output from the source buffer so i could keep sending forms to it. It seemed strange to me that inf-clojure obviously knows which buffer you want to clear but demands you move point to that buffer before clearing it.

bbatsov: Got it. Well, I guess it'd be nice if we had some helper function like `current-repl` or `with-current-repl` to abstract this away, as I can assume this is going to be needed in other cases as well.

PR #188, opened 2021-03-14
Handle newlines between forms for `inf-clojure-eval-buffer`

dpsutton:
This sends the contents of the buffer. However, newlines cause the
prompt to be returned. You can see this with even a regular clojure
repl:
```bash
❯❯❯ clojure
Clojure 1.10.2
user=>
user=>
user=>
```
(note using `clj` has rlwrap which hides a bit of this so make sure to
use `clojure`).

But what we can do is make sure to transform
```clojure
(defn foo [] ...)

(defn bar [] ...)
```

into

```clojure
(defn foo [] ...)
(defn bar [] ...)
```

So that the newlines don't trigger more repl prompts. Real world usage below:

Before:
```clojure
parse=>
nil
parse=> parse=> nil
parse=> parse=> #'parse/data
parse=> parse=> #'parse/parse-where
parse=> parse=> #'parse/keywords
parse=> parse=> #'parse/tokenize
parse=>
parse=> #'parse/parse
parse=> parse=> #'parse/translate-where
parse=> parse=> nil
parse=> parse=> #'parse/query
parse=> parse=> #'parse/query-test
parse=> parse=> #'parse/bonus-points-test
parse=>
```

After:
```clojure
user=>
nil
parse=> nil
parse=> #'parse/data
parse=> #'parse/parse-where
parse=> #'parse/keywords
parse=>
parse=> #'parse/parse
parse=> #'parse/translate-where
parse=> nil
parse=> #'parse/query
parse=> #'parse/query-test
parse=> #'parse/bonus-points-test
parse=>
```

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -683,16 +683,40 @@ HOST is the host the process is running on, PORT is where it's listening."
   (interactive "shost: \nnport: ")
   (inf-clojure (cons host port)))
 
+(defun inf-clojure--forms-without-newlines (str)
+  "Remove newlines between toplevel forms.
+STR is a string of contents to be evaluated.  When sending
+multiple forms to a socket repl, each newline triggers a prompt.
+So we replace all newlines between top level forms but not inside
+of forms."
+  (condition-case nil
+      (with-temp-buffer
+        (progn
+          (clojurec-mode)
+          (insert str)
+          (whitespace-cleanup)
+          (goto-char (point-min))
+          (while (not (eobp))
+            (while (looking-at "\n")
+              (delete-char 1))
+            (unless (eobp)
+              (clojure-forward-logical-sexp))
+            (unless (eobp)
+              (forward-char)))
+          (buffer-substring-no-properties (point-min) (point-max))))
+    (scan-error str)))
+
 (defun inf-clojure-eval-region (start end &optional and-go)
   "Send the current region to the inferior Clojure process.
 Sends substring between START and END.  Prefix argument AND-GO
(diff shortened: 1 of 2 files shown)


inf-clojure.el
```diff
@@ -688,11 +688,27 @@ HOST is the host the process is running on, PORT is where it's listening."
...
-  ;; drops newlines at the end of the region
-  (let ((str (replace-regexp-in-string
-              "[\n]+\\'" ""
-              (buffer-substring-no-properties start end))))
-    (inf-clojure--send-string (inf-clojure-proc) str))
+  (let* ((str (buffer-substring-no-properties start end))
+         ;; newlines over a socket repl between top level forms cause
+         ;; a prompt to be returned. so here we dump the region into a
+         ;; temp buffer, and delete all newlines between the forms
+         (formatted (condition-case nil
```
bbatsov, line 695: That's seems a bit too complicated to me. Can't we just use the built-in `delete-blank-lines` or something like this?

bbatsov: Actually, I see you tried to kill the blank lines only between top-level forms. That seems reasonable, but we should probably move this to some helper function, so it's clearer what it does (and we can add a test or two about it).

dpsutton: will do tonight thanks!

dpsutton:
@bbatsov the change is pretty big but look at the last commit. There were a bunch of poorly formatted tests so i just knocked those out.

bbatsov:
Looks good. Thanks!

PR #189, opened 2021-03-14
Font lock on repl inserts

dpsutton:
Just a simple call to font lock the repl when inserting

image


bbatsov:
Won't it be better to just enable font-locking for everything in the inf-clojure buffer? (perhaps controlled by a defcustom) That's just a matter of doing something like:

`  (clojure-font-lock-setup)`

to the major mode init code.

bbatsov:
My bad, seems we've already done this. Hmm, I wonder what's stripping the font-locking, which normally should be preserved anyways.

PR #190, opened 2021-03-20
Function to select repls as the current active connection

dpsutton:
Helpful little function to ease the use case of multiple connections. There's a global connection that all commands use: `inf-clojure-buffer`. Changing this allows using different connections. This function just sets that to current buffer if in a repl, shows a list of live process inf-clojure repls if not, and if given a prefix, always shows a list so you can switch even if you're currently in a repl. The repl type is in a defvar-local `inf-clojure-repl-type` so that doesn't need to be set or re-inferred.

Renaming buffers with projects and repl types makes this quite nice, and the dead simple way it works means that things can be evaluated from any buffer (project irrespective) and no edge cases like last focused buffer or anything like that.

image

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -277,7 +277,8 @@ mode.  Default is whitespace followed by 0 or 1 single-letter colon-keyword
 (defun inf-clojure--modeline-info ()
   "Return modeline info.
 Either \"not connected\" or \"repl-type: buffer-name\""
-  (if (bufferp inf-clojure-buffer)
+  (if (and (bufferp inf-clojure-buffer)
+           (buffer-live-p inf-clojure-buffer))
       (with-current-buffer inf-clojure-buffer
         (format "%s: %s" inf-clojure-repl-type (buffer-name (current-buffer))))
     "not connected"))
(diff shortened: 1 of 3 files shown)


inf-clojure.el
```diff
@@ -199,6 +199,29 @@ has been found.  See also variable `inf-clojure-buffer'."
...
         (error "No Clojure subprocess; see variable `inf-clojure-buffer'"))))
 
+(defun inf-clojure-set-repl (always-ask)
+  "Set an inf clojure buffer as the active repl.
+If in a repl already, use that unless a prefix is used (or
+ALWAYS-ASK).  Otherwise get a list of all active inf-clojure
+repls and offer a choice.  Recommended to rename buffers as they
+are created with `rename-buffer`."
+  (interactive "P")
+  (cl-flet ((inf-clojure-repl-p () (and (derived-mode-p 'inf-clojure-mode)
```
bbatsov, line 209: I'd suggest having this and a `inf-clojure-repls` fucntion to be top-level functions. I can imagine they'd be useful in some other situations as well.

bbatsov:
Nice little change indeed. Might also be a good idea to update the modeline for `inf-clojure-minor-mode` to display the active REPL (similar to what CIDER does). That's another simple change.

arichiardi:
Looks like I will have to try inf-clojure again 😃 Good idea here!

dpsutton:
image

image

bbatsov:
Nice!

PR #191, opened 2021-03-22
Comint adds *'s to name:

dpsutton:
docstring of `make-comint`:
The name of the buffer is made by surrounding NAME with `*'s.

Otherwise you end up in a fundamental mode buffer unconnected to the
process.

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -750,9 +750,11 @@ process buffer for a list of commands.)"
                                           nil
                                           'confirm-after-completion))))
   (let* ((project-dir (clojure-project-dir))
-         (repl-buffer-name (if project-dir
-                               (format "*inf-clojure %s*" (inf-clojure--project-name project-dir))
-                             "*inf-clojure*")))
+         (process-buffer-name (if project-dir
+                                  (format "inf-clojure %s" (inf-clojure--project-name project-dir))
+                                "inf-clojure"))
+         ;; comint adds the asterisks to both sides
+         (repl-buffer-name (format "*%s*" process-buffer-name)))
     ;; Create a new comint buffer if needed
     (unless (comint-check-proc repl-buffer-name)
       ;; run the new process in the project's root when in a project folder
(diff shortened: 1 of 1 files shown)


inf-clojure.el
```diff
@@ -766,7 +768,7 @@ process buffer for a list of commands.)"
                            (inf-clojure--prompt-repl-type))))
         (message "Starting Clojure REPL via `%s'..." cmd)
         (with-current-buffer (apply #'make-comint
-                                    "inf-clojure" (car cmdlist) nil (cdr cmdlist))
+                                    process-buffer-name (car cmdlist) nil (cdr cmdlist))
```
bbatsov, line 771: Are you sure that's needed? At first I thought so myself, but you might have noticed that I just reverted it, before you added it. :D I don't think that the process name has to be unique or derived from the buffer name. After all - it was always the same until now and this worked just fine.

dpsutton: without this change inf-clojure would make a fundamental mode buffer I end up in a fundamental buffer named `*inf-clojure build-your-own-logic-engine*` and the correct inf-clojure repl buffer is hidden and named `**inf-clojure build-your-own-logic-engine**`
image

dpsutton: prior to this change the string "inf-clojure" was passed in, and the buffer name we used was "*inf-clojure*" for this reason

bbatsov: e.g. that's from IELM:

image

Seems to me there's no need for the names to match, but I might be missing something. I no longer remember comint's APIs well. :D

bbatsov: Actually, you're right. That's the only place where we pass the buffer name anyways. :D I wonder what I was thinking when I deleted this. :D

dpsutton: i suspect that's the disambiguation that they do when they do match. ie, the same name goes in and it adds a 1 on the end. I suspect the best thing to do is actually start up the repl and then rename it `repl-projec-name`

bbatsov:
Redundant, my ass https://github.com/clojure-emacs/inf-clojure/commit/9841b38dde980789b3b5d13479c3b299fa3fcbd8 😆

inf-clojure.el
bbatsov: You're probably right, I didn't dwell on this much. I was also thinking of making the naming template configurable in case someone wants different names by default.

bbatsov:
You're welcome! In the morning I felt inspired and I decided to do a bit of work on `inf-clojure` myself. Your PR about the default REPL reminded me that there's still plenty of low-hanging fruit that we can easily tackle.

PR #197, opened 2022-03-01
Adds an option to disable eldoc

dpsutton:
Eldoc is quite nice and puts the function signatures in the
minibuffer. But it does this at a cost. Since inf-clojure only uses a
single connection (currently at least) the commands interrupt the
values of `*1`, `*2`, etc. Further, this can lead to multiple prompts
appearing in the repl buffer.

```clojure
user=> user=> (map inc (range 4))
(1 2 3 4)
user=> user=> *1
nil
user=>
```

`user` appears multiple times, and then `*1` has been bound to the
result of getting arglists

```clojure
user=> (+ 1 1)
2
user=> *1
([f] [f coll] [f c1 c2] [f c1 c2 c3] [f c1 c2 c3 & colls])
```

The multiple prompts is quite annoying when inserting forms into the
repl.

Note that `*1` is oftentimes `nil` because eldoc just runs `(try (:arglists (meta _whatever-your-cursor-is-on)) (catch nil))` whenever your cursor stays on something for too long. This is also not throwing away too much functionality since you can easily get the docstring in the repl by `C-c C-v` (`inf-clojure-show-var-documentation`)

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -384,7 +384,7 @@ mode line entirely."
   :type 'sexp
   :risky t)
 
-(defvar inf-clojure-eldoc-enabledp t
+(defcustom inf-clojure-eldoc-enabledp t
   "Var that allows disabling `eldoc-mode` in `inf-clojure`.
 
 Set to `nil` to disable eldoc.  Eldoc can be quite useful by
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -1403,10 +1403,18 @@ Return the number of nested sexp the point was over or after."
           (setq inf-clojure-eldoc-last-symbol (cons thing arglists))
           arglists)))))
 
+(defvar inf-clojure-eldoc-enabledp t
+  "Var that allows disabling `eldoc-mode` in `inf-clojure`.
+
+Set to `nil` to disable eldoc.  Eldoc can be quite useful by
+displaying funciton signatures in the modeline, but can also
+cause multiple prompts to appear and mess with `*1`, `*2`, etc.")
+
 (defun inf-clojure-eldoc ()
   "Backend function for eldoc to show argument list in the echo area."
   ;; todo: this never gets unset once connected and is a lie
-  (when (and (inf-clojure-connected-p)
+  (when (and inf-clojure-eldoc-enabledp
+             (inf-clojure-connected-p)
              ;; don't clobber an error message in the minibuffer
              (not (member last-command '(next-error previous-error))))
     (let* ((info (inf-clojure-eldoc-info-in-current-sexp))
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -1416,6 +1416,7 @@ Return the number of nested sexp the point was over or after."
   "Backend function for eldoc to show argument list in the echo area."
   ;; todo: this never gets unset once connected and is a lie
   (when (and (inf-clojure-connected-p)
+             inf-clojure-eldoc-enabledp
              ;; don't clobber an error message in the minibuffer
              (not (member last-command '(next-error previous-error))))
     (let* ((info (inf-clojure-eldoc-info-in-current-sexp))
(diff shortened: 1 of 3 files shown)


bbatsov:
Can't people just turn-off `eldoc-mode` locally? (e.g. using a hook)

dpsutton:
I think you are right that this is not the way to go forward. A hook to turn off eldoc mode helps for newly created buffers but is annoying for any that already exist. But perhaps that's not so bad. Let me think on it for a bit. I'm wondering if other modes in the buffer use eldoc so that turning eldoc mode completely off is heavy handed.

Andre0991:
>  I'm wondering if other modes in the buffer use eldoc so that turning eldoc mode completely off is heavy handed.

[`eglot`](https://github.com/joaotavora/eglot/) (LSP client) uses eldoc.

Andre0991:
By the way, as an inspiration, `eglot` has an variable for not interfering with other modes:

```
`eglot-stay-out-of: List of Emacs features that Eglot shouldn't automatically try to manage on users' behalf. Useful when you need non-LSP Flymake or Company backends. See docstring for examples.
```

I wonder if `inf-clojure` could do the same for `eldoc`.

inf-clojure.el
```diff
@@ -384,6 +384,13 @@ mode line entirely."
   :type 'sexp
   :risky t)
 
+(defvar inf-clojure-eldoc-enabledp t
+  "Var that allows disabling `eldoc-mode` in `inf-clojure`.
+
+Set to `nil` to disable eldoc.  Eldoc can be quite useful by
+displaying funciton signatures in the modeline, but can also
```
bbatsov, line 391: Btw, this disclaimer is true for every command that does some code evaluation on the side. :-)

dpsutton: That's true, but all others are user driver. If you refresh a namespace or call doc, you understand that `*1` and others will have been moved because you caused an evaluation. Eldoc is special because unbeknownst to the user those vars will change value. It makes using them very frustrating.

dpsutton:
I've updated it to be a bit more sensible i think. Now I just prohibit the eldoc setup when it is not enabled. This allows other modes to use eldoc and seems quite nice in practice for me.

inf-clojure.el
bbatsov: Also - there's a typo in `function`.

inf-clojure.el
```diff
@@ -631,7 +639,8 @@ to continue it."
   (setq mode-line-process '(":%s"))
   (clojure-mode-variables)
   (clojure-font-lock-setup)
-  (inf-clojure-eldoc-setup)
+  (when inf-clojure-eldoc-enabledp
```
bbatsov, line 642: I'm ok with this, but it might be better to make the eldoc integration a no-op instead, so that people won't have to disable/enable inf-clojure-mode for the new config to be taken into account.

bbatsov:
I've added a couple of small feedback remarks. This also needs to be documented somewhere in the README.

inf-clojure.el
dpsutton: That was originally my solution, but that still leads to some problems. Eldoc is initialized as follows:

```emacs-lisp
(defun inf-clojure-eldoc-setup ()
  "Turn on eldoc mode in the current buffer."
  (setq-local eldoc-documentation-function #'inf-clojure-eldoc)
  (apply #'eldoc-add-command inf-clojure-extra-eldoc-commands))
```

Since it clobbers any existing `eldoc-documentation-function`, once eldoc is initialized in a buffer there's no good way to go back to a clean slate since any previous eldoc function is gone. I'll add the `enabledp` function to the eldoc function so it would prevent more eldocs, but it cannot restore the previous eldoc unfortunately.

dpsutton:
All addressed I believe. Thanks for the thoughtful review @bbatsov !

bbatsov:
You're welcome! Thanks for tackling this.

PR #199, opened 2022-04-29
Remove reader conditional from Clojure repl type arglists default config

JasonKDarby:
The existing configuration causes Clojure repls to fail with `java.lang.RuntimeException: Conditional read not allowed` as soon as arglists are requested.

I ran into this while using an embedded nrepl configured with transport/tty, here's [a dummy project you can use](https://github.com/JasonKDarby/dummy-embedded-nrepl-tty) to try it out.

It's not clear to me if this is sacrificing functionality that nrepl transport/tty doesn't implement or if it never worked.

Thanks for inf-clojure!

-----------------

Thanks!

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -775,7 +775,7 @@ process buffer for a list of commands.)"
       (let ((default-directory (or project-dir default-directory))
             (cmdlist (if (consp cmd)
                          (list cmd)
-                       (split-string cmd)))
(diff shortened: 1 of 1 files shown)


JasonKDarby:
It looks like this undoes https://github.com/clojure-emacs/inf-clojure/pull/164, maybe a better fix exists.

Issue #168, opened 2019-11-07
would be nice if C-c C-z in REPL would switch back to source buffer


vigilancetech-com:
# Expected behavior
C-c C-z switches back
## Actual behavior
displays C-c C-z
## Steps to reproduce the problem
1. bring up a clojurescript source file
2. inf-clojure-connect
3. put whatever you have in to connect (e.g. localhost RETURN 5555 RETURN)
4. hit C-c C-z (cursor switches to REPL buffer)
5. hit C-c C-z again (cursor stays where it is but prints out C-c C-z)

### inf-clojure version information

2.2.0-snapshot

### Lein/Boot version

using lumo 1.10.1

### Emacs version

26.2

### Operating system

Windows 10

bbatsov:
Yep, that'd be pretty simple to do.

onetom:
This mapping would solve this specific situation, I guess:
```elisp
(define-key inf-clojure-mode-map (kbd "C-c C-z") #'other-window)
```
However, `inf-clojure-switch-to-repl` has some logic for jumping across frames too, based on the `pop-up-frames` custom variable, so this might not work if frames are involved too.

bbatsov:
I think the simplest option would be to remember the current "location" somewhere in `inf-clojure-switch-to-repl` and then add a matching command that just jumps to that "saved location".

bbatsov:
(or you can check for the most recently visited `clojure-mode` buffer as an alternative)

bbatsov:
Fixed in #200.

PR #200, opened 2022-05-04
Added function for switching back from an inf-clojure buffer

JasonKDarby:
This addresses https://github.com/clojure-emacs/inf-clojure/issues/168 by creating `inf-clojure--recent-buffer` to keep track of the buffer the repl was switched to from.  It's set on both `inf-clojure` and `inf-clojure-switch-to-repl` which hopefully means it can never be unintentionally `nil`.  If `inf-clojure-switch-to-recent-buffer` is called and `inf-clojure--recent-buffer` IS `nil` the message `"inf-clojure: No recent buffer known."` is sent.

This only tracks the most recent buffer among all repls, it will not manage multiple references in the case of multiple repls.

I'm by no means an elisp expert, if there are improvements to be made I'm happy to hear them.  I'm gonna try to take on more changes, I love having an opportunity to contribute!

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
+* [#168](https://github.com/clojure-emacs/inf-clojure/pull/197): Helper function `inf-clojure-switch-to-recent-buffer` to select the last buffer an inf-clojure process buffer was swapped to from.
 * [#187](https://github.com/clojure-emacs/inf-clojure/pull/197): Defcustom `inf-clojure-enable-eldoc` to disable eldoc interaction.
 
 ## 3.1.0 (2021-07-23)
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -816,7 +816,9 @@ process buffer for a list of commands.)"
   "Connect to a running socket REPL server via `inf-clojure'.
 HOST is the host the process is running on, PORT is where it's listening."
   (interactive "shost: \nnport: ")
-  (inf-clojure (cons host port)))
+  (progn
+    (setq inf-clojure--recent-buffer (current-buffer))
+    (inf-clojure (cons host port))))
 
 (defun inf-clojure--forms-without-newlines (str)
   "Remove newlines between toplevel forms.
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -702,7 +702,7 @@ to continue it."
          ;; that already has the window in it.
          (or pop-up-frames
              (get-buffer-window to-buffer t))))
-    (pop-to-buffer to-buffer)))
+    (pop-to-buffer to-buffer '(display-buffer-reuse-window . ()))))
 
 (defun inf-clojure-switch-to-repl (eob-p)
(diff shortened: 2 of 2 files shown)


bbatsov:
Looks good to me overall. Just mention this in the changelog and I we can merge this.

JasonKDarby:
Thanks!  I did already, if I should change it let me know.

I did notice I forgot to update `inf-clojure--recent-buffer` on `inf-clojure-connect`, I can add that later.

JasonKDarby:
As far as I know it's good to go, let me know if there's anything else.

dpsutton:
This attempts to track a buffer you came from. This seems inherently weak as there are only certain times we mark the buffer we came from. Seems like most times it will be nil or stale. Can we do something like what CIDER does and look through the `(buffer-list)` for the most recent buffer that is in a mode that descends from `clojure-mode`?

For reference, CIDER's version:
```emacs-lisp
(defun cider-switch-to-last-clojure-buffer ()
  "Switch to the last Clojure buffer.
The default keybinding for this command is
the same as variable `cider-switch-to-repl-buffer',
so that it is very convenient to jump between a
Clojure buffer and the REPL buffer."
  (interactive)
  (if (derived-mode-p 'cider-repl-mode)
      (let* ((a-buf)
             (the-buf (let ((repl-type (cider-repl-type-for-buffer)))
                        (seq-find (lambda (b)
                                    (unless (with-current-buffer b (derived-mode-p 'cider-repl-mode))
                                      (when-let* ((type (cider-repl-type-for-buffer b)))
                                        (unless a-buf
                                          (setq a-buf b))
                                        (or (eq type 'multi)
                                            (eq type repl-type)))))
                                  (buffer-list)))))
        (if-let* ((buf (or the-buf a-buf)))
            (if cider-repl-display-in-current-window
                (pop-to-buffer-same-window buf)
              (pop-to-buffer buf))
          (user-error "No Clojure buffer found")))
    (user-error "Not in a CIDER REPL buffer")))
```

But the gist is `(seq-find clojure-mode-predicate (buffer-list))`

JasonKDarby:
@dpsutton I was only intending to update it when it was swapped with C-c C-z / `inf-clojure-switch-to-repl`, is this to say the last buffer when swapped to by any means (ex C-xo)?

dpsutton:
@JasonKDarby exactly. The current implementation would take you back to the last buffer that you used `C-c C-z` in, which might be a while ago, landing you in a strange buffer. It seems to me that the ideal workflow is a quick way to bounce between the repl and a source buffer, regardless of how you got there.

JasonKDarby:
@dpsutton I think this works, I've tried it out with two windows.
Before passing `'(display-buffer-reuse-window . ())` to `pop-to-buffer` I kept having this issue where it would change back to the correct buffer but would always do it in the 'first' window.  I'm not sure why that wasn't happening before but it looks like it's fixed now.

Let me know if there's anything else to do on this.  I love inf-clojure.

dpsutton:
thanks @JasonKDarby . this is a great addition.

PR #201, opened 2022-05-13
Addressed prompt duplication by printing nothing when response is prompt by itself

JasonKDarby:
This change is intended to fix the problem of the prompt being printed in line with a previously printed prompt.  The issue manifests as a prompt that looks like `user=> user=>` and has the potential to print many more times than that.

It happens when an evaluation takes long enough for shorter requests to finish before the long one does.  This has commonly come up for me when running tests or invoking `load-file` while completions are configured to execute through the inf-clojure process.

This pr addresses the fix in 2 steps.

First, ALL non-blanked responses are prepended by newline, as opposed to only those fulfilling `(string-prefix-p "inf-clojure-" (symbol-name (or this-command last-command)))`.  At a minimum this makes sure multiple results can't show up on the same line, which also helps to prevent the prompt from getting misaligned.  I wasn't entirely certain what was going on with the `string-prefix-p`, in my testing it didn't appear to affect inf-clojure- evals, though they were somehow being filtered anyway.

Second, `inf-clojure--previous-response-ended-with-prompt` keeps track of whether the previous response ended with a prompt.  If it was, and if we get only a prompt in the next request, we can replace the response with an empty string, which will leave the point at the previous prompt and appear to the user as if there was no response.

I don't consider this ready to merge, I really want to use it for a while and make sure it's enough.  So far it appears to be working pretty well.

-----------------

Thanks!

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -676,14 +676,22 @@ to continue it."
   "Remove subprompts from STRING."
   (replace-regexp-in-string inf-clojure-subprompt "" string))
 
+(defvar inf-clojure--previous-response-ended-with-prompt nil)
+
 (defun inf-clojure-preoutput-filter (str)
   "Preprocess the output STR from interactive commands."
   (inf-clojure--log-string str "<-RES----")
-  (cond
-   ((string-prefix-p "inf-clojure-" (symbol-name (or this-command last-command)))
-    ;; Remove subprompts and prepend a newline to the output string
-    (inf-clojure-chomp (concat "\n" (inf-clojure-remove-subprompts str))))
-   (t str)))
+  (let* ((last-line (car (last (split-string str "\n"))))
+         (response-was-one-line (string= last-line str))
+         (last-line-was-prompt (string-match inf-clojure-prompt last-line))
+         (filtered (if (and inf-clojure--previous-response-ended-with-prompt
+                            response-was-one-line
+                            last-line-was-prompt)
+                       ;; Prevent prompts from stacking in the same line.
+                       ""
+                     (inf-clojure-chomp (concat "\n" str)))))
+    (set 'inf-clojure--previous-response-ended-with-prompt last-line-was-prompt)
+    filtered))
 
 (defun inf-clojure-clear-repl-buffer ()
   "Clear the REPL buffer."
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -684,12 +684,19 @@ to continue it."
   (let* ((last-line (car (last (split-string str "\n"))))
          (response-was-one-line (string= last-line str))
          (last-line-was-prompt (string-match inf-clojure-prompt last-line))
-         (filtered (if (and inf-clojure--previous-response-ended-with-prompt
-                            response-was-one-line
-                            last-line-was-prompt)
-                       ;; Prevent prompts from stacking in the same line.
-                       ""
-                     (inf-clojure-chomp (concat "\n" str)))))
+         (filtered (cond
+                    ;; Just a prompt.
+                    ((and inf-clojure--previous-response-ended-with-prompt
+                          response-was-one-line
+                          last-line-was-prompt)
+                     "")
+                    ;; Not just a prompt, not the beginning of a response, don't insert unnecessary lines.
+                    ;; This should be a continuation of a previous response.
+                    ((not inf-clojure--previous-response-ended-with-prompt)
+                     str)
+                    ;; This should be the beginning of a response.
+                    (t
+                     (concat "\n" str)))))
     (set 'inf-clojure--previous-response-ended-with-prompt last-line-was-prompt)
     filtered))
 
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -676,8 +676,6 @@ to continue it."
   "Remove subprompts from STRING."
   (replace-regexp-in-string inf-clojure-subprompt "" string))
 
-(defvar inf-clojure--previous-response-ended-with-prompt nil)
-
 (defun inf-clojure-preoutput-filter (str)
   "Preprocess the output STR from interactive commands."
   (inf-clojure--log-string str "<-RES----")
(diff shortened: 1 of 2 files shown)


inf-clojure.el
```diff
@@ -676,14 +676,22 @@ to continue it."
...
 (defun inf-clojure-preoutput-filter (str)
   "Preprocess the output STR from interactive commands."
   (inf-clojure--log-string str "<-RES----")
-  (cond
-   ((string-prefix-p "inf-clojure-" (symbol-name (or this-command last-command)))
-    ;; Remove subprompts and prepend a newline to the output string
-    (inf-clojure-chomp (concat "\n" (inf-clojure-remove-subprompts str))))
-   (t str)))
+  (let* ((last-line (car (last (split-string str "\n"))))
+         (response-was-one-line (string= last-line str))
```
JasonKDarby, line 685: I would appreciate some elisp feedback here.  I could get the same result by checking the length of the split str but I'm not sure if there's a scenario where this is a more valid thing to do.  I mostly just did it because I didn't like the idea of another let binding for the split str.

inf-clojure.el
```diff
@@ -676,14 +676,22 @@ to continue it."
...
+         (last-line-was-prompt (string-match inf-clojure-prompt last-line))
+         (filtered (if (and inf-clojure--previous-response-ended-with-prompt
+                            response-was-one-line
+                            last-line-was-prompt)
+                       ;; Prevent prompts from stacking in the same line.
+                       ""
+                     (inf-clojure-chomp (concat "\n" str)))))
```
JasonKDarby, line 692: This has the effect of putting an extra empty line after manually entered forms (because enter adds \n) but I don't know any way to associate a response with a command.  I guess the other way to solve it would be to scrub the \n from the command after it's entered, not sure how practical that is.

dpsutton:
I think there's a bug when first `inf-clojure-connect` that it does not print the first repl.

Repro:
- start a socket repl on port 6000
- `inf-clojure-connect` `localhost` `6000`
Buffer looks blank with no indication it is ready. But if you type in it it will evaluate and then print a prompt:

```clojure
(+ 1 1)
2
user=> 
```

JasonKDarby:
I think I've seen that when the repl doesn't print anything except the prompt on connect.  I'm not sure why the prompt would be removed when `inf-clojure--previous-response-ended-with-prompt` is `nil` but I'll do some debugging.

JasonKDarby:
@dpsutton so the issue comes up when you connect to a repl that only shows the prompt 2 times.  This is because `inf-clojure--previous-response-ended-with-prompt` gets set during the first connection, and during the second connection causes the starting prompt to get scrubbed.  I can fix this by resetting `inf-clojure--previous-response-ended-with-prompt` to nil during repl connections.

This will still be an issue in the case of multiple simultaneous repl connections.  Because `inf-clojure--previous-response-ended-with-prompt` tracks repl state and is only a single reference there isn't a particularly easy way to address it.  I could work on tracking state of all potential repl connections, maybe just as an alist of repl buffer names to their individual `inf-clojure--previous-response-ended-with-prompt`s but how would I know when a repl buffer changed names (according to the guide for using multiple repls with inf-clojure)?

dpsutton:
> This will still be an issue in the case of multiple simultaneous repl connections. Because inf-clojure--previous-response-ended-with-prompt tracks repl state and is only a single reference there isn't a particularly easy way to address it. I could work on tracking state of all potential repl connections, maybe just as an alist of repl buffer names to their individual inf-clojure--previous-response-ended-with-prompts but how would I know when a repl buffer changed names (according to the guide for using multiple repls with inf-clojure)?

I think emacs has a native construct for this: `defvar-local`.
```
defvar-local is a Lisp macro in ‘subr.el’.

(defvar-local VAR VAL &optional DOCSTRING)

  Probably introduced at or before Emacs version 24.3.

Define VAR as a buffer-local variable with default value VAL.
Like ‘defvar’ but additionally marks the variable as being automatically
buffer-local wherever it is set.
```

JasonKDarby:
Surprisingly straightforward!

inf-clojure.el
```diff
@@ -803,6 +816,7 @@ process buffer for a list of commands.)"
           (inf-clojure-mode)
           (set-syntax-table clojure-mode-syntax-table)
           (setq-local inf-clojure-repl-type repl-type)
+          (defvar-local inf-clojure--previous-response-ended-with-prompt nil)
```
bbatsov, line 819: It seems to me this should where all the other `defvar-locals` are, not inside the mode definition itself.

inf-clojure.el
bbatsov: Probably it makes a bit more sense first to check if something was multiline and only then split it.

inf-clojure.el
```diff
@@ -682,11 +682,24 @@ to continue it."
...
+                          last-line-was-prompt)
+                     "")
+                    ;; Not just a prompt, not the beginning of a response, don't insert unnecessary lines.
+                    ;; This should be a continuation of a previous response.
+                    ((not inf-clojure--previous-response-ended-with-prompt)
+                     str)
+                    ;; This should be the beginning of a response.
+                    (t
+                     (concat "\n" str)))))
+    (set 'inf-clojure--previous-response-ended-with-prompt last-line-was-prompt)
```
bbatsov, line 701: It's much more common to use `setq` in Elisp.

inf-clojure.el
```diff
@@ -682,11 +682,24 @@ to continue it."
 (defun inf-clojure-preoutput-filter (str)
   "Preprocess the output STR from interactive commands."
   (inf-clojure--log-string str "<-RES----")
-  (cond
-   ((string-prefix-p "inf-clojure-" (symbol-name (or this-command last-command)))
-    ;; Remove subprompts and prepend a newline to the output string
-    (inf-clojure-chomp (concat "\n" (inf-clojure-remove-subprompts str))))
-   (t str)))
+  (let* ((last-line (car (last (split-string str "\n"))))
```
bbatsov, line 685: I guess you can also add some high-level description of the need for this pre-processing in the docstring or above the `let*`.

inf-clojure.el
```diff
@@ -682,11 +682,24 @@ to continue it."
...
+         (response-was-one-line (string= last-line str))
+         (last-line-was-prompt (string-match inf-clojure-prompt last-line))
+         (filtered (cond
```
bbatsov, line 688: Perhaps `filtered-output` or `processed-output`. `filtered` looks a bit weird to me.

bbatsov:
I finally took a closer look at the PR and I've added a bit of small feedback here and there.

bbatsov:
@JasonKDarby ping :-)

JasonKDarby:
!

Thanks for the ping, I'll look back into this

JasonKDarby:
I'm going to close this PR but leave my branch available.  I haven't been able to get back to this lately, I'll resubmit once I do.

PR #203, opened 2022-07-21
Add ClojureCLR support

E-A-Griffin:
Added support for ClojureCLR repl via `lein-clr`. Updated README to reflect changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## master (unreleased)
+* [#202](https://github.com/clojure-emacs/inf-clojure/issues/202): Add ClojureCLR support.
 
 ## 3.2.0 (2022-07-15)
 
@@ -123,4 +124,3 @@
 ### New Features
 
 * [#22](https://github.com/clojure-emacs/inf-clojure/pull/22): Add ElDoc support.
-* [#202](https://github.com/clojure-emacs/inf-clojure/issues/202): Add ClojureCLR support.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -123,3 +123,4 @@
 ### New Features
 
 * [#22](https://github.com/clojure-emacs/inf-clojure/pull/22): Add ElDoc support.
+* [#202](https://github.com/clojure-emacs/inf-clojure/issues/202): Add ClojureCLR support.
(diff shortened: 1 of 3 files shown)


inf-clojure.el
```diff
@@ -75,6 +75,7 @@
                                     (boot . "boot repl")
                                     (clojure . "clojure")
                                     (cljs . "clojure -m cljs.main -r")
+                                    (cljr . "lein clr repl")
```
bbatsov, line 78: Might be better to name this `lein-cljr` in case we decide to add more ways to start a cljr repl down the road.

bbatsov:
Please, mention this in the changelog as well.

bbatsov:
Btw, I've also noticed that `lein-clr` hasn't been updated in 7 years, but I guess it still gets the job done.

E-A-Griffin:
> Btw, I've also noticed that `lein-clr` hasn't been updated in 7 years, but I guess it still gets the job done.

Yeah I noticed that as well. There's at least one feature I think it should be updated to include (dealing with dependency management for Clojure/ClojureScript/ClojureCLR projects where some dependencies aren't compatible with all three variants of Clojure). I may reach out to the repo owner and see if it would be appropriate to move the repo to something like https://github.com/clj-commons. I'd be willing to try to contribute to its maintenance though I wouldn't want to take on that responsibility solely.

inf-clojure.el
E-A-Griffin: I completely agree, especially with the hope that more modern tools for `cljr` repls become available soon.

bbatsov:
Sounds like a good idea to move the project to `clj-commons`.

E-A-Griffin:
Let me know if these changes address all concerns or if further changes are still needed.

CHANGELOG.md
```diff
@@ -123,3 +123,4 @@
 ### New Features
 
 * [#22](https://github.com/clojure-emacs/inf-clojure/pull/22): Add ElDoc support.
+* [#202](https://github.com/clojure-emacs/inf-clojure/issues/202): Add ClojureCLR support.
```
bbatsov, line 126: That should be under the `master` section.

bbatsov:
@E-A-Griffin Just move the changelog entry and we're good to go.

E-A-Griffin:
> @E-A-Griffin Just move the changelog entry and we're good to go.

Awesome, should be good now then.

PR #204, opened 2022-08-06
Ensure repl scrolls on insert

dpsutton:
When inserting forms into the repl, the buffer doesn't always scroll to
show the new input and result. This has been fixed in CIDER in a similar
way so bringing the same trick here.

https://github.com/clojure-emacs/cider/pull/2590

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -920,6 +920,8 @@ Prefix argument AND-GO means switch to the Clojure buffer afterwards."
   "Insert FORM into process and evaluate.
 Indent FORM.  FORM is expected to have been trimmed."
   (let ((clojure-process (inf-clojure-proc)))
+    ;; ensure the repl buffer scrolls. See similar fix in CIDER:
+    ;; https://github.com/clojure-emacs/cider/pull/2590
     (with-selected-window (or (get-buffer-window inf-clojure-buffer)
                               (selected-window))
       (with-current-buffer (process-buffer clojure-process)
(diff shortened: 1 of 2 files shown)


inf-clojure.el
```diff
@@ -920,16 +920,19 @@ Prefix argument AND-GO means switch to the Clojure buffer afterwards."
...
-      (comint-goto-process-mark)
-      (let ((beginning (point)))
-        (insert (format "%s" form))
-        (let ((end (point)))
-          (goto-char beginning)
-          (indent-sexp end)
-          ;; font-lock the inserted code
-          (font-lock-ensure beginning end)))
-      (comint-send-input t t))))
+    (with-selected-window (or (get-buffer-window inf-clojure-buffer)
```
bbatsov, line 923: What is this for?

dpsutton: I linked the CIDER PR that does the same thing. That points to https://stackoverflow.com/questions/13530025/emacs-scroll-automatically-when-inserting-text . If the point is not visible the `(goto (point-max))` code won't actually scroll the buffer. 

This change ensures that inserting into the repl will scroll the repl so you can see what you inserted and the result. Before this it only worked if the point was at the end and visible.

bbatsov: Got it - admittedly I didn't read the PRs description, only the title. :D Might be good to mention this in some comment as well.

dpsutton: on it! thanks!

PR #207, opened 2022-11-13
Use comint-mode-map as parent map instead

Inc0n:
Whilst playing with inf-clojure-mode-map, some previous binds are overriden, because copy-keymap is used instead of set-parent-map. This will help for that.


bbatsov:
Thanks!

Inc0n:
A small change, but a convenient one :)

PR #208, opened 2023-02-24
Display message after setting repl

jjttjj:
Displays a message after setting the repl.

When manually setting a repl with `inf-clojure-set-repl`, there is no visual indication that the repl has successfully been changed. This has confused me a few times in the past so I added a message.

This was [briefly discussed on slack](https://clojurians.slack.com/archives/C099W16KZ/p1672414905867309)

-----------------

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## master (unreleased)
 * [#202](https://github.com/clojure-emacs/inf-clojure/issues/202): Add ClojureCLR support.
 * [#204](https://github.com/clojure-emacs/inf-clojure/issues/204): Scroll repl buffer on insert commands
+* [#208](https://github.com/clojure-emacs/inf-clojure/pull/208) Display message after setting repl.
 
 
 ## 3.2.1 (2022-07-22)
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -281,14 +281,17 @@ ALWAYS-ASK).  Otherwise get a list of all active inf-clojure
 REPLS and offer a choice.  It's recommended to rename REPL
 buffers after they are created with `rename-buffer'."
   (interactive "P")
-  (if (and (not always-ask)
-           (inf-clojure-repl-p))
-      (setq inf-clojure-buffer (current-buffer))
-    (let ((repl-buffers (inf-clojure-repls)))
-     (if (> (length repl-buffers) 0)
-         (when-let ((repl-buffer (completing-read "Select default REPL: " repl-buffers nil t)))
-           (setq inf-clojure-buffer (get-buffer repl-buffer)))
-       (user-error "No buffers have an inf-clojure process")))))
+  (let ((new-repl-buffer
+         (if (and (not always-ask)
+                  (inf-clojure-repl-p))
+             (setq inf-clojure-buffer (current-buffer))
+           (let ((repl-buffers (inf-clojure-repls)))
+             (if (> (length repl-buffers) 0)
+                 (when-let ((repl-buffer (completing-read "Select default REPL: " repl-buffers nil t)))
+                   (setq inf-clojure-buffer (get-buffer repl-buffer)))
+               (user-error "No buffers have an inf-clojure process"))))))
+    (when new-repl-buffer
+      (message "Current inf-clojure REPL set to %s" new-repl-buffer))))
 
 (defvar inf-clojure--repl-type-lock nil
   "Global lock for protecting against proc filter race conditions.
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -274,6 +274,16 @@ Checks the mode and that there is a live process."
         (push (buffer-name b) repl-buffers)))
     repl-buffers))
 
+(defun inf-clojure--prompt-repl-buffer (prompt)
+  "Prompt the user to select an inf-clojure repl buffer.
+PROMPT is a string to prompt the user.
+Returns nil when no buffer is selected."
+  (let ((repl-buffers (inf-clojure-repls)))
+    (if (> (length repl-buffers) 0)
+        (when-let ((repl-buffer (completing-read prompt repl-buffers nil t)))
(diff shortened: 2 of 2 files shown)


inf-clojure.el
```diff
@@ -285,10 +285,11 @@ buffers after they are created with `rename-buffer'."
...
     (let ((repl-buffers (inf-clojure-repls)))
-     (if (> (length repl-buffers) 0)
-         (when-let ((repl-buffer (completing-read "Select default REPL: " repl-buffers nil t)))
-           (setq inf-clojure-buffer (get-buffer repl-buffer)))
-       (user-error "No buffers have an inf-clojure process")))))
+      (if (> (length repl-buffers) 0)
+          (when-let ((repl-buffer (completing-read "Select default REPL: " repl-buffers nil t)))
+            (setq inf-clojure-buffer (get-buffer repl-buffer)))
+        (user-error "No buffers have an inf-clojure process"))))
+  (message "inf-clojure repl set to %s" inf-clojure-buffer))
```
bbatsov, line 292: Probably it'd be better to use something like "Current/active inf-clojure REPL set to ...".

dpsutton:
thanks for the addition. Great usability addition. Once you update to @bbatsov 's suggestion for the message we'll merge.

inf-clojure.el
jjttjj: Thanks for the input, I made the change!

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## master (unreleased)
 * [#202](https://github.com/clojure-emacs/inf-clojure/issues/202): Add ClojureCLR support.
 * [#204](https://github.com/clojure-emacs/inf-clojure/issues/204): Scroll repl buffer on insert commands
+* Display message after setting repl.
```
dpsutton, line 6: you should add a link here 
> * [#208](https://github.com/clojure-emacs/inf-clojure/pull/208) Display message after setting repl with `inf-clojure-set-repl`

inf-clojure.el
```diff
@@ -285,10 +285,11 @@ buffers after they are created with `rename-buffer'."
...
     (let ((repl-buffers (inf-clojure-repls)))
-     (if (> (length repl-buffers) 0)
-         (when-let ((repl-buffer (completing-read "Select default REPL: " repl-buffers nil t)))
-           (setq inf-clojure-buffer (get-buffer repl-buffer)))
-       (user-error "No buffers have an inf-clojure process")))))
+      (if (> (length repl-buffers) 0)
+          (when-let ((repl-buffer (completing-read "Select default REPL: " repl-buffers nil t)))
+            (setq inf-clojure-buffer (get-buffer repl-buffer)))
+        (user-error "No buffers have an inf-clojure process"))))
+  (message "Current inf-clojure REPL set to %s" inf-clojure-buffer))
```
dpsutton, line 292: Also, should this `message` move this to be inside of the `when-let` which sets `inf-clojure-buffer`?

jjttjj: You're right, I missed that possible outcome.
I changed it to only display the message after a successful change. Not confident in my style choices here, I could also just repeat the call to `message` in both success branches separately, let me know if you have a preference. 
https://github.com/jjttjj/inf-clojure/blob/04b967f1e60adb79cd84d580268f7af00021eb71/inf-clojure.el#L284-L294

dpsutton: I don't have any strong opinions. I lean towards having the `message` and the `setq` in the same body so they work in tandem. But it doesn't really matter.

inf-clojure.el
```diff
@@ -281,17 +281,17 @@ ALWAYS-ASK).  Otherwise get a list of all active inf-clojure
...
-                  (inf-clojure-repl-p))
-             (setq inf-clojure-buffer (current-buffer))
-           (let ((repl-buffers (inf-clojure-repls)))
-             (if (> (length repl-buffers) 0)
-                 (when-let ((repl-buffer (completing-read "Select default REPL: " repl-buffers nil t)))
-                   (setq inf-clojure-buffer (get-buffer repl-buffer)))
-               (user-error "No buffers have an inf-clojure process"))))))
-    (when new-repl-buffer
-      (message "Current inf-clojure REPL set to %s" new-repl-buffer))))
+  (if (and (not always-ask)
```
bbatsov, line 284: The nest ifs make it a bit hard to follow the code, so some comments at the beginning of each branch might be useful.

inf-clojure.el
```diff
@@ -281,17 +281,17 @@ ALWAYS-ASK).  Otherwise get a list of all active inf-clojure
...
+           (inf-clojure-repl-p))
+      (progn
+        (setq inf-clojure-buffer (current-buffer))
+        (message "Current inf-clojure REPL set to %s" inf-clojure-buffer))
```
bbatsov, line 288: To reduce this message duplication you can just return the new-repl buffer at the end and set it outside the `if`s with a single message there. You can even the the code looking for the new repl to be a separate private function, so this would read a bit better. I'd prefer such less imperative code structure.

jjttjj:
After trying a few arrangements, I went with extracting a function to just handle the "select repl" prompt. I believe this could also be reused to refactor [this](https://github.com/clojure-emacs/inf-clojure/blob/e5ce3839835b9b561fca5810f43f413c96c197d9/inf-clojure.el#L1484-L1501) if you would like. 
(Sorry if this little change has required a disproportionate amount of attention)

bbatsov:
> After trying a few arrangements, I went with extracting a function to just handle the "select repl" prompt. I believe [this](https://github.com/clojure-emacs/inf-clojure/blob/e5ce3839835b9b561fca5810f43f413c96c197d9/inf-clojure.el#L1484-L1501) could also be reused to refactor this if you would like.

Yeah, it'd be nice to remove the code duplication, so a follow-up PR will be appreciated. 

>  (Sorry if this little change has required a disproportionate amount of attention)

No worries! The code is better in the end and that is all that matters.

Issue #211, opened 2023-04-07
Symbol's value as variable is void: inf-clojure-custom-repl-name


jakub-stastny:
## Expected behavior

`inf-clojure-connect` connects me to `localhost` `5555` (which I pass into the fn).

## Actual behavior

`inf-clojure-connect` asks me for `host` and `port` and then crashes with `Symbol's value as variable is void: inf-clojure-custom-repl-name`.

## Steps to reproduce the problem

- Start a Clojure socket REPL in shell.
- Then do `inf-clojure-connect`.
- It'll ask `host`, `port` and finally it crashes before asking REPL type.

## Environment & Version information

- M2 MB Air

### inf-clojure version information

`inf-clojure-20230326.1946`

### Emacs version

GNU Emacs 28.2

### Operating system

Latest macOS.

jakub-stastny:
How to fix in the meantime: 

```lisp
(defvar inf-clojure-custom-repl-name nil)
(defvar inf-clojure-socket-repl-type nil)
```

Kudos to @dpsutton.

PR #212, opened 2023-04-07
Cleanup socket repl startup.

dpsutton:
Fixes: https://github.com/clojure-emacs/inf-clojure/issues/211

**How to use:**

To use one of the defaults defined in
`inf-clojure-socket-repl-startup-forms`, just `m-x inf-clojure-socket-repl` and choose your startup.

To add your own custom startup

```emacs-lisp
(setq inf-clojure-custom-startup
      "clojure -J-Dclojure.server.repl=\"{:port %d :accept clojure.core.server/repl}\" -A:grepl")
```

This will prompt you for a repl type so it's useful to also include

```emacs-lisp
(setq inf-clojure-custom-repl-type 'clojure)
```

Or set these in a .dir-locals.el file for each project with

```emacs-lisp
;;; Directory Local Variables
;;; For more information see (info "(emacs) Directory Variables")

((nil . ((inf-clojure-custom-startup . "clojure -J-Dclojure.server.repl=\"{:port %d :accept clojure.core.server/repl}\" -A:grepl")
         (inf-clojure-custom-repl-type . clojure))))
```

**things i changed**
- added `(defvar inf-clojure-custom-repl-name nil)`. This was originally used but never defined so socket startup and inf-connect would fail. This also was not honored and now controls the repl buffer name.
- removed `inf-clojure-socket-repl-type`. This was also used and never defined so it also caused startup to fail. But we already have a notion of `inf-clojure-custom-repl-type` so we can just reuse that. There's nothing special about it being the socket repl type for connect.
- simplified the buffer startup mechanism. Lots of `let` bindings
- added `inf-clojure--suppress-connect-message-p`. Ostensibly private, others can use it. But two things are sending messages: how we start the clojure process, and that we connected to the clojure process. I think how we start the clojure process is the far more important one as it includes port information, aliases, etc.

cc: @jakub-stastny @mikepjb

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -518,6 +518,12 @@ Should be a symbol that is a key in `inf-clojure-repl-features'."
                  (const :tag "babashka" babashka)
                  (const :tag "determine at startup" nil)))
 
+(defvar inf-clojure-custom-repl-name nil
+  "A string to be used as the repl buffer name.")
+
+(defvar inf-clojure--suppress-connect-message-p nil
+  "A boolean of whether to suppress the message on connect.")
+
 (defun inf-clojure--whole-comment-line-p (string)
   "Return non-nil iff STRING is a whole line semicolon comment."
   (string-match-p "^\s*;" string))
@@ -833,12 +839,12 @@ process buffer for a list of commands.)"
             (cmdlist (if (consp cmd)
                          (list cmd)
                        (split-string-and-unquote cmd)))
-            (repl-type (or inf-clojure-socket-repl-type
-                           (unless prefix-arg
+            (repl-type (or (unless prefix-arg
                              inf-clojure-custom-repl-type)
-                        (car (rassoc cmd inf-clojure-startup-forms))
-                        (inf-clojure--prompt-repl-type))))
-        (message "Starting Clojure REPL via `%s'..." cmd)
+                           (car (rassoc cmd inf-clojure-startup-forms))
+                           (inf-clojure--prompt-repl-type))))
+        (unless inf-clojure--suppress-connect-message-p
+          (message "Starting Clojure REPL via `%s'..." cmd))
         (with-current-buffer (apply #'make-comint
                                     process-buffer-name (car cmdlist) nil (cdr cmdlist))
           (inf-clojure-mode)
@@ -849,7 +855,8 @@ process buffer for a list of commands.)"
     (setq inf-clojure-buffer (get-buffer repl-buffer-name))
     (if inf-clojure-repl-use-same-window
         (pop-to-buffer-same-window repl-buffer-name)
-      (pop-to-buffer repl-buffer-name))))
+      (pop-to-buffer repl-buffer-name))
+    repl-buffer-name))
 
 ;;;###autoload
 (defun inf-clojure-connect (host port)
@@ -882,7 +889,6 @@ OUTPUT is the latest data received from the process"
         (insert output)))
     (let ((prompt-displayed (string-match inf-clojure-prompt output)))
       (when prompt-displayed
-	(message (format "Socket REPL startup detected for %s" (process-name process)))
 	(with-current-buffer server-buffer
 	  (when inf-clojure-socket-callback
 	    (funcall inf-clojure-socket-callback)))))))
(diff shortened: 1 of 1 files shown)


inf-clojure.el
```diff
@@ -917,11 +923,11 @@ If left as nil a random port will be selected between 5500-6000."
...
-  "Start a socket REPL server and connect to it via `inf-clojure'.
-CMD is the command line used to start the socket REPL, if this
-isn't provided you will be prompted to select from the defaults
-provided in `inf-clojure-socket-repl-startup-forms' or
-`inf-clojure-custom-startup' if this is defined."
+  "Start a socket REPL server and connects to it via `inf-clojure-connect'.
+CMD is the command line instruction used to start the socket
+REPL.  It should be a string with \"%d\" in it to take a random
+port.  Set `inf-clojure-custom-startup' or choose from the
+defaults provided in `inf-clojure-socket-repl-startup-forms'"
```
bbatsov, line 930: This should end with a `.`.

inf-clojure.el
```diff
@@ -833,12 +839,12 @@ process buffer for a list of commands.)"
...
-            (repl-type (or inf-clojure-socket-repl-type
-                           (unless prefix-arg
+            (repl-type (or (unless prefix-arg
                              inf-clojure-custom-repl-type)
-                        (car (rassoc cmd inf-clojure-startup-forms))
-                        (inf-clojure--prompt-repl-type))))
-        (message "Starting Clojure REPL via `%s'..." cmd)
+                           (car (rassoc cmd inf-clojure-startup-forms))
+                           (inf-clojure--prompt-repl-type))))
+        (unless inf-clojure--suppress-connect-message-p
```
bbatsov, line 846: I'm not a big fan of dynamic vars and I'd probably remove this altogether and replace it with some optional param for the containing function.

bbatsov:
Seems I didn't pay close enough attention the PR that introduced those changes. My bad! Your PR looks good to me overall, apart from my small comments. Also I've noticed the following in the CI log:

```
[00:04.474]  
             In toplevel form:
             inf-clojure.el:614:1: Warning: custom-declare-variable `inf-clojure-auto-mode'
                 docstring wider than 80 characters
[00:04.501]  
             In inf-clojure-socket-repl-sentinel:
             inf-clojure.el:896:50: Warning: Unused lexical argument `event'
```

Seems those warnings got obscured by the test that has been failing for a while. Probably we should comment it out at this point.

inf-clojure.el
dpsutton: updated to have an optional passed arg.

dpsutton:
I want to fix up all of the lint errors but I really don't understand the one that is complaining about the regex `"^\\( *#_\\|[^=> \n]+\\)=> *"`. Also one complains that we have to depend on emacs 28.3 if we use eldoc which seems obviously wrong.

inf-clojure.el
```diff
@@ -609,7 +609,9 @@ This should usually be a combination of `inf-clojure-prompt' and
   :package-version '(inf-clojure . "2.0.0"))
 
 (defcustom inf-clojure-auto-mode t
-  "When non-nil, automatically enable inf-clojure-minor-mode for all Clojure buffers."
+  "Automatically enable inf-clojure-minor-mode.
+All buffers will automatically be in `inf-clojure-minor-mode'
+unless set to nil.."
```
bbatsov, line 614: .. -> .

dpsutton:
full lint run:
```
inf-clojure on  master [$?] on ☁️  metabase-query
❯ eldev lint
Bootstrapping Eldev for Emacs 28.2 from MELPA Stable...

inf-clojure.el:1142: Keycode C-u embedded in doc string.  Use \\<mapvar> & \\[command] instead
Found 1 warning in file ‘inf-clojure.el’
[1/1] Installing package ‘package-lint’ (0.17) from ‘melpa-stable’...
inf-clojure.el:1:0: warning: Including "Emacs" in the package summary is usually redundant.
inf-clojure.el:1103:0: error: `inf-clojure-prev-l/c-dir/file' contains a non-standard separator `/', use hyphens instead (see Elisp Coding Conventions).
inf-clojure.el:1374:10: error: You should depend on (emacs "28.1") if you need `eldoc'.
Found 3 warnings in file ‘inf-clojure.el’
[1/2] Installing package ‘xr’ (1.23) from ‘gnu’...
[2/2] Installing package ‘relint’ (1.22) from ‘gnu’...
inf-clojure.el:1505:49: In call to skip-chars-backward: Duplicated character ‘[’ (pos 16)
  "^[] \"'`><,;|&{()[@\\^]"
   .................^
inf-clojure.el:1505:49: In call to skip-chars-backward: Duplicated character ‘]’ (pos 20)
  "^[] \"'`><,;|&{()[@\\^]"
   ......................^
Found 2 warnings in file ‘inf-clojure.el’
Linting produced warnings
```

inf-clojure.el
```diff
@@ -609,7 +609,9 @@ This should usually be a combination of `inf-clojure-prompt' and
   :package-version '(inf-clojure . "2.0.0"))
 
 (defcustom inf-clojure-auto-mode t
-  "When non-nil, automatically enable inf-clojure-minor-mode for all Clojure buffers."
+  "Automatically enable inf-clojure-minor-mode.
+All buffers will automatically be in `inf-clojure-minor-mode'
```
bbatsov, line 613: The updated wording sounds a bit confusing to me. Perhaps you still want to mention this is for Clojure buffers.

dpsutton: Good call. I wanted to fix it but emacs is quite annoying about the first line being short and then having a subsequent line. I had to make two duplicative sentences instead of just a single one because of like 3 characters. That and the two spaces after periods rule always annoy me. :)

PR #213, opened 2023-08-30
Skip empty prompts.

dop:

-----------------

Thanks!

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -722,6 +722,10 @@ If `comint-use-prompt-regexp' is nil (the default), \\[comint-insert-input] on
   "Remove subprompts from STRING."
   (replace-regexp-in-string inf-clojure-subprompt "" string))
 
+(defun inf-clojure--empty-prompt-p (str)
+  "Returns t if STR is an empty prompt, e.g. \"user=> \"."
+  (eq 0 (string-match (concat inf-clojure-prompt "$") str)))
+
 (defun inf-clojure-preoutput-filter (str)
   "Preprocess the output STR from interactive commands."
   (inf-clojure--log-string str "<-RES----")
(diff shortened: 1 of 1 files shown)


dpsutton:
@dop thanks for the PR!

I think I'm following the code easily enough. But do you think you could add a PR description? It would be nice if we knew where the prompts are coming from. Is this useful mostly when running a subprocess? Or when connected on a socket-repl? 

Can you add some tests? And also we'll want to include this nice functionality in the changelog.

dop:
Oh my, I did not intend to open a PR. Sorry for the noise. I'll let you know if I actually understand the issue.

dpsutton:
@dop well I'm glad you're thinking about the issue. Feel free to discuss here if you want to brainstorm on it.

dop:
@dpsutton I did some investigation and came up with this solution:

```diff
@@ -1293,7 +1286,7 @@ defun inf-clojure--process-response
 will end at (point-max) in the results buffer.  It cuts out the
 output from and including the `inf-clojure-prompt`."
   (let ((redirect-buffer-name inf-clojure--redirect-buffer-name)
-        (sanitized-command (inf-clojure--sanitize-command command)))
+        (sanitized-command (string-trim command)))
     (when (not (string-empty-p sanitized-command))
       (inf-clojure--log-string command "----CMD->")
       (with-current-buffer (inf-clojure--get-redirect-buffer)
```

Why?

`inf-clojure--sanitize-command` appends newline to the command, but so does comint, effectively doing this:

```
$ echo -n "(inc 1)\n\n" | clojure
Clojure 1.11.1
user=> 2
user=> user=>
```

Source of these prompt duplication for me comes from trying to figure arglist of a symbol at point. It uses comint redirect feature. At this point I have only the speculation, but since comint redirect is looking for prompt as the end of output it sometimes would receive `2\nuser=>` while in redirect mode leaving second `user=>` to be passed to regular process filter thus ending in comint buffer. This makes sense to me after reading how `comint-redirect-completed` and `accept-process-output` work.

PR #215, opened 2023-11-20
Add clojure-ts-mode support

Andre0991:
(draft PR, will update this description as I get things working correctly)

---

-----------------

Thanks!


inf-clojure.el
```diff
@@ -64,7 +64,8 @@
 ;;; Code:
 
 (require 'comint)
-(require 'clojure-mode)
+(require 'clojure-mode nil 'no-error)
```
bbatsov, line 67: I think the mode still has a hard dependency on `clojure-mode`, as it uses functions from it internally. That's why you can't really make this optional.

bbatsov: (all the functions that start with `clojure-` are coming from there)

bbatsov:
You might also want to check for inspiration a similar PR for CIDER https://github.com/clojure-emacs/cider/pull/3461

inf-clojure.el
Andre0991: Oops. Makes sense. I just reverted this line to just `(require 'clojure-mode)`

Andre0991:
Cool, I'll implement it.

bbatsov:
@Andre0991 Any updates?

PR #217, opened 2025-05-24
Add `clojure-ts-mode` support

rrudakov:
#215 has no activity for more than 2 years, so I decided to create a new one.

Initially I planned to add a new `inf-clojure-preferred-major-mode` variable, but then I discovered `inf-clojure-source-modes` and I think it's a good idea to use it for this feature.

`clojure-ts-mode` is used in the most places if selected as a preferred major mode, except a few places where required functions are missing. I left comments in the code.

-----------------

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,10 @@ You can also add the following to your Emacs config to enable
 
 ```emacs-lisp
 (add-hook 'clojure-mode-hook #'inf-clojure-minor-mode)
+
+;; or if you're a `clojure-ts-mode' user:
+
+(add-hook 'clojure-ts-mode-hook #'inf-clojure-minor-mode)
 ```
 
 **Warning:** Don't enable `inf-clojure-minor-mode` and `cider-mode` at the same time. They
@@ -215,6 +219,18 @@ If you want to update a specific form there is a function
 
 ```emacs-lisp
 (inf-clojure-update-feature 'clojure 'completion "(incomplete.core/completions \"%s\")")
+```
+
+### `clojure-ts-mode` support
+
+`inf-clojure` will try to use `clojure-ts-mode` by default if it's
+available with fallback to `clojure-mode`.
+
+If you want to use `inf-clojure` with `clojure-mode` exclusively, you
+can set it to:
+
+```emacs-lisp
+(setopt inf-clojure-source-modes '(clojure-mode))
 ```
 
 #### Caveats
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -65,11 +65,13 @@
 
 (require 'comint)
 (require 'clojure-mode)
+(require 'clojure-ts-mode nil :no-error)
 (require 'eldoc)
 (require 'thingatpt)
 (require 'ansi-color)
 (require 'cl-lib)
 (require 'subr-x)
+(require 'project)
 
 (defvar inf-clojure-startup-forms '((lein . "lein repl")
                                     (boot . "boot repl")
@@ -193,10 +195,10 @@ either `setq-local` or an entry in `.dir-locals.el`." )
 MULTIPLE PROCESS SUPPORT
 ===========================================================================
 To run multiple Clojure processes, you start the first up
-with \\[inf-clojure].  It will be in a buffer named `*inf-clojure*'.
+with \\[inf-clojure].  It will be in a buffer named *inf-clojure*.
 Rename this buffer with \\[rename-buffer].  You may now start up a new
 process with another \\[inf-clojure].  It will be in a new buffer,
-named `*inf-clojure*'.  You can switch between the different process
+named *inf-clojure*.  You can switch between the different process
 buffers with \\[switch-to-buffer].
 
 Commands that send text from source buffers to Clojure processes --
@@ -205,7 +207,7 @@ process to send to, when you have more than one Clojure process around.  This
 is determined by the global variable `inf-clojure-buffer'.  Suppose you
 have three inferior Clojures running:
     Buffer              Process
-    foo                 inf-clojure
+    foo                 `inf-clojure'
     bar                 inf-clojure<2>
     *inf-clojure*     inf-clojure<3>
 If you do a \\[inf-clojure-eval-defun] command on some Clojure source code,
@@ -269,7 +271,7 @@ has been found.  See also variable `inf-clojure-buffer'."
         (error "No Clojure subprocess; see variable `inf-clojure-buffer'"))))
 
 (defun inf-clojure-repl-p (&optional buf)
-  "Indicates if BUF is an inf-clojure REPL.
+  "Indicates if BUF is an `inf-clojure' REPL.
 If BUF is nil then defaults to the current buffer.
 Checks the mode and that there is a live process."
   (let ((buf (or buf (current-buffer))))
@@ -278,7 +280,7 @@ Checks the mode and that there is a live process."
          (process-live-p (get-buffer-process buf)))))
 
 (defun inf-clojure-repls ()
-  "Return a list of all inf-clojure REPL buffers."
+  "Return a list of all `inf-clojure' REPL buffers."
   (let (repl-buffers)
     (dolist (b (buffer-list))
       (when (inf-clojure-repl-p b)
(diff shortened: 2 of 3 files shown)


inf-clojure.el
```diff
@@ -986,8 +1016,9 @@ of forms."
             (while (looking-at "\n")
               (delete-char 1))
             (unless (eobp)
-              (clojure-forward-logical-sexp))
-            (unless (eobp)
+              ;; There is no special API for that in
```
dpsutton, line 1019: You removed the `(clojure-forward-logical-sexp)` call but added a comment that you probably keep it in. Comment and action are at odds here.

rrudakov: Yep, removed by accident. Will bring it back. Thanks!

inf-clojure.el
```diff
@@ -205,7 +206,7 @@ process to send to, when you have more than one Clojure process around.  This
 is determined by the global variable `inf-clojure-buffer'.  Suppose you
 have three inferior Clojures running:
     Buffer              Process
-    foo                 inf-clojure
+    foo                 `inf-clojure'
```
dpsutton, line 209: why did you add the quotes to this line but not the line below it? Adding these formatting changes into this PR also clouds where the work is.

rrudakov: One of the requirements for PRs is to run `M-x checkdoc`. It forced me to add quotes here and there and fix some other things. I can revert it, but then there will be warnings.

README.md
```diff
@@ -217,6 +221,22 @@ If you want to update a specific form there is a function
 (inf-clojure-update-feature 'clojure 'completion "(incomplete.core/completions \"%s\")")
 ```
 
+### `clojure-ts-mode` support
+
+To use `inf-clojure` with `clojure-ts-mode`, you can adjust the
+following variable:
+
+```emacs-lisp
+(setopt inf-clojure-source-modes '(clojure-ts-mode clojure-mode))
```
bbatsov, line 230: Shouldn't we just make this the default?

inf-clojure.el
```diff
@@ -688,8 +708,12 @@ If `comint-use-prompt-regexp' is nil (the default), \\[comint-insert-input] on
   (setq comint-input-sender 'inf-clojure--send-string)
   (setq comint-prompt-regexp inf-clojure-comint-prompt-regexp)
   (setq mode-line-process '(":%s"))
-  (clojure-mode-variables)
-  (clojure-font-lock-setup)
+  ;; We don't need to call it in Clojure buffers.
+  (unless (inf-clojure--clojure-buffer-p)
```
bbatsov, line 712: I don't get the need for this check - after all that's the init of the REPL's major mode.

inf-clojure.el
```diff
@@ -978,6 +1006,8 @@ of forms."
   (condition-case nil
       (with-temp-buffer
         (progn
+          ;; TODO: Should it be adjusted for `clojure-ts-mode'
```
bbatsov, line 1009: Yeah, probably. I guess we can infer the major mode to prefer from what's currently active in the open clojure buffers (e.g. we take the first Clojure buffer and check it's mode or something along those lines). I had suggested to add an option for a preferred mode mostly to avoid the need for this inference and to easily pick the right functions from both modes, but there are different ways to approach this.

inf-clojure.el
```diff
@@ -688,8 +708,12 @@ If `comint-use-prompt-regexp' is nil (the default), \\[comint-insert-input] on
   (setq comint-input-sender 'inf-clojure--send-string)
   (setq comint-prompt-regexp inf-clojure-comint-prompt-regexp)
   (setq mode-line-process '(":%s"))
-  (clojure-mode-variables)
-  (clojure-font-lock-setup)
+  ;; We don't need to call it in Clojure buffers.
+  (unless (inf-clojure--clojure-buffer-p)
+    ;; Using Tree-sitter based syntax highlighting in comint buffer is
```
bbatsov, line 713: Ah, this was a research item on my todo list. What's the blocker there currently? (I assumed the parser could be constrained to the current prompt or something along those lines)

inf-clojure.el
rrudakov: Indeed, my bad. For some reason I thought it's used for Clojure source buffers as well. Will revert it.

inf-clojure.el
rrudakov: Well, either we should expose some API from `clojure-ts-mode` or bring some functions from `treesit.el` to `inf-clojure`.

inf-clojure.el
rrudakov: Later in this function `(clojure-forward-logical-sexp)` is used, which might not work as expected in `clojure-ts-mode`, so for now probably it would be better to keep `clojurec-mode` here. BTW, is there a specific reason why `clojurec-mode` is used and not just `clojure-mode`?

inf-clojure.el
rrudakov: Reverted. Thanks!

inf-clojure.el
```diff
@@ -826,6 +850,8 @@ process buffer for a list of commands.)"
                                           (mapcar #'cdr inf-clojure-startup-forms)
                                           nil
                                           'confirm-after-completion))))
+  ;; TODO: Currently there is no `clojure-project-dir' analogue in
```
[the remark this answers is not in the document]
bbatsov: Yeah, I know it doesn't, but as you noted its behavior can be tweaked. Or we can just inline the project logic here as well. I had added it to `clojure-mode` in the past to avoid having it in both CIDER and inf-clojure, but technically speaking it doesn't fit very well a typical major mode. (as it's not directly used anywhere in the mode)

inf-clojure.el
bbatsov: I think that'd be perfectly fine, as ideally we should get to the point where there's no hard dep on `clojure-mode` in both CIDER and inf-clojure.

inf-clojure.el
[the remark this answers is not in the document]
rrudakov: This works, I've just tested:

```emacs-lisp
(defun inf-clojure--project-dir ()
  (let ((project-vc-extra-root-markers '("project.clj"      ; Leiningen
                                         "build.boot"       ; Boot
                                         "build.gradle"     ; Gradle
                                         "build.gradle.kts" ; Gradle
                                         "deps.edn"         ; Clojure CLI (a.k.a. tools.deps)
                                         "shadow-cljs.edn"  ; shadow-cljs
                                         "bb.edn"           ; babashka
                                         "nbb.edn"          ; nbb
                                         "basilisp.edn"     ; Basilisp (Python)
                                         )))
    (project-root (project-current))))
```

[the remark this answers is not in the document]
bbatsov: Let's go with it then.

inf-clojure.el
bbatsov: I don't remember, but if I had to guess - probably `clojure-mode` didn't behave properly in `cljc` files. At any rate - you can just go with `clojure-mode` and `clojure-ts-mode` and we'll know soon enough if this causes any problems.

inf-clojure.el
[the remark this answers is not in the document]
rrudakov: Done.

README.md
rrudakov: Please let me know if I should proceed with changing the default.

bbatsov: Yeah, you can proceed.

inf-clojure.el
rrudakov: Sounds good. But this will require some extra research, I tried to copy some font-lock-* related calls from `treesit.el` and it didn't work out. I'll add a note to handle it separately.

Also it looks like this package should be compatible with at least Emacs 26.2, I believer treesit.el is not available in this version yet.

README.md
rrudakov: Done. Also adjusted the README.

inf-clojure.el
rrudakov: FTR, I made it work by calling the following functions:

```emacs-lisp
(treesit-parser-create 'clojure)
(clojure-ts-mode-variables nil nil)
(treesit-major-mode-setup)
```

image

The more history in the REPL buffer, the bigger is the syntax tree. Not sure if that might be a problem.

bbatsov: Probably it won't be a bit issue in practice. 

P.S. It'd be nice to apply something similar to CIDER's REPL as well.

bbatsov:
@rrudakov I've noticed that the CI was not running and I fixed it today. Seems your changes introduced a few warnings https://app.circleci.com/pipelines/github/clojure-emacs/inf-clojure/123/workflows/de4bff6a-64d8-458d-8ba7-e337b2ea9eff/jobs/413

Nothing major, but I wanted to mention it so we won't forget about them.

PR #218, opened 2025-05-25
Fix CI failures

rrudakov:
- Looks like `project-root` function was introduced in Emacs 28.1. So tests fail on Emacs 27.
- Linter complains about duplicated square brackets in the regex `"^[] \"'`><,;|&{()[@\\^]"`. I'm not sure it's an error, maybe one of the pairs should be escaped?

-----------------

Thanks!


rrudakov:
@bbatsov what's the best way to trigger CI?

bbatsov:
I'm guess for you'd it be easiest to just push changes here. I'm not sure you can run jobs directly from the console https://app.circleci.com/pipelines/github/clojure-emacs/inf-clojure

bbatsov:
> Linter complains about duplicated square brackets in the regex "^[] \"'><,;|&{()[@\^]"`. I'm not sure it's an error, maybe one of the pairs should be escaped?

Looking more closely at this code - it doesn't even seem that's a regexp (it's passed only to `skip-chars-backward` whose arg is not a regexp). But for some reason that's confusing the regexp linter.

rrudakov:
One of the failing tests is:

```emacs-lisp
(it "computes no bounds for point directly after a break expression"
    (ict-with-assess-buffers
     ((a (insert "@")))
     (with-current-buffer a
       (expect
        (ict-bounds-string (inf-clojure-completion-bounds-of-expr-at-point))
        :not :to-be nil))))
```

I honestly don't really understand what it's about and Git history doesn't give me much details either. Any tips would be welcome :)

bbatsov:
Ah, probably it's actually supposed to be empty string when there are no relevant characters.

rrudakov:
Looking at the code I would assume that returning `nil` from this function is OK. It's used in 2 places and the result is always checked for `nil`. Maybe the test is incorrect after all?

```emacs-lisp
(defun inf-clojure-completion-bounds-of-expr-at-point ()
  "Return bounds of expression at point to complete."
  (when (not (memq (char-syntax (following-char)) '(?w ?_)))
    (save-excursion
      (let* ((end (point))
             (skipped-back (skip-chars-backward inf-clojure-clojure-expr-break-chars))
             (start (+ end skipped-back))
             (chars (or (thing-at-point 'symbol)
                        (inf-clojure--kw-to-symbol (buffer-substring start end)))))
        (when (> (length chars) 0)
          (let ((first-char (substring-no-properties chars 0 1)))
            (when (string-match-p "[^0-9]" first-char)
              (cons (point) end))))))))

(defun inf-clojure-completion-expr-at-point ()
  "Return expression at point to complete."
  (let ((bounds (inf-clojure-completion-bounds-of-expr-at-point)))
    (and bounds
         (buffer-substring (car bounds) (cdr bounds)))))

(defun inf-clojure-completion-at-point ()
  "Retrieve the list of completions and prompt the user.
Returns the selected completion or nil."
  (let ((bounds (inf-clojure-completion-bounds-of-expr-at-point)))
    (when (and bounds (inf-clojure-get-feature (inf-clojure-proc) 'completion 'no-error))
      (list (car bounds) (cdr bounds)
            (if (fboundp 'completion-table-with-cache)
                (completion-table-with-cache #'inf-clojure-completions)
              (completion-table-dynamic #'inf-clojure-completions))))))
```

rrudakov:
> > Looks like project-root function was introduced in Emacs 28.1. So tests fail on Emacs 27.
> 
> I see. I guess we can require Emacs 28. It's fairly old by now anyways.

Should I remove Emacs 27 from CI config?

bbatsov:
I'm fine with your proposal on both points.

rrudakov:
Sorry for the noise again. I've done some digging in the Git history and found out that this test was introduced in this PR: https://github.com/clojure-emacs/inf-clojure/pull/129

The original regex was `" \t\n\"\'`><,;|&{()[]@\\^"` (commit https://github.com/clojure-emacs/inf-clojure/pull/129/commits/74a6fac31914d32f524e52894d95ee45085dbaf9). If I change the regex to this, the test passes, it returns the boundaries of `@` character.

Then in another commit (https://github.com/clojure-emacs/inf-clojure/pull/129/commits/dbe2839dd2f1f14b4466be9cfcd220608630806f) it was changed to the current value and this test stopped working. I'm a bit hesitant to change the regex. Maybe we should use `rx` to simplify it?

bbatsov:
I don't think `rx` would help, as this is not a regexp ( the docstring is wrong) - the only usage is here              

(skipped-back (skip-chars-backward inf-clojure-clojure-expr-break-chars))

and this function takes a list of chars (a string). I guess we can update the tests and the wrong var description for and call it a day.

rrudakov:
> I don't think `rx` would help, as this is not a regexp ( the docstring is wrong) - the only usage is here
> 
> (skipped-back (skip-chars-backward inf-clojure-clojure-expr-break-chars))
> 
> and this function takes a list of chars (a string). I guess we can update the tests and the wrong var description for and call it a day.

Haha, that's what I've just figured out as well. But from the original PR I assume it should match `@`, and currently it doesn't. I'll try to make it work.

rrudakov:
I decided to update the test to expect `nil`. The string, expected by `skip-chars-backward` is not as simple as it seems:

```
STRING is like the inside of a ‘[...]’ in a regular expression
except that ‘]’ is never special and ‘\’ quotes ‘^’, ‘-’ or ‘\’
 (but not at the end of a range; quoting is never needed there).
Thus, with arg "a-zA-Z", this skips letters stopping before first nonletter.
With arg "^a-zA-Z", skips nonletters stopping before first letter.
Char classes, e.g. ‘[:alpha:]’, are supported.

Returns the distance traveled, either zero or positive.
```

Moving characters around breaks tests, so there is still warning about duplicated brakets and I don't really know how to fix it, because it looks like a valid case to have one "special" `]` and one non-special in the same string.

All the tests are fixed now, @bbatsov it's ready for review.

bbatsov:
> I guess it tries to lint it as an actual regex, but it's not aware that characters in this string have special meaning.

I made the same discovery earlier today when I tried to fix this myself. Not sure how to instruct the linter to ignore this, but we'll figure it out later.

PR #222, opened 2026-02-20
Fix preoutput filter to handle chunked comint output

### LINKED ISSUE #136: Comint pre output filter handling is wrong
While digging some other issue, I discovered that the way we are using `comint-preoutput-filter-functions` might be off. An example output:

```
----CMD->
"(lumo.repl/doc defn)\n"
<-RES----
"-------------------------"
<-RES----
"\ncljs.core/defn\n([name doc-string? attr-map? [params*] prepost-map? body] [name doc-string? attr-map? ([params*] prepost-map? body) + attr-map?])\n  Same as (def name (core/fn [params* ] exprs*)) or (def\n    name (core/fn ([params* ] exprs*)+)) with any doc-string or attrs added\n    to the var metadata. prepost-map defines a map with optional keys\n    :pre and :post that contain collections of pre or post conditions.\nnil\ncljs.user=> cljs.user=> "
```

It might be for a couple of reasons, and the above might happen in `lumo` only`, but it seems that `inf-clojure-preoutput-filter` is called more than once.

More to dig.

## Expected behavior
Ideally it should filter the aggregated response from the repl.

## Actual behavior
The current split produces two prompts (one per filtered response) like:

```
-------------------------

cljs.core/defn
([name doc-string? attr-map? [params*] prepost-map? body] [name doc-string? attr-map? ([params*] prepost-map? body) + attr-map?])
  Same as (def name (core/fn [params* ] exprs*)) or (def
    name (core/fn ([params* ] exprs*)+)) with any doc-string or attrs added
    to the var metadata. prepost-map defines a map with optional keys
    :pre and :post that contain collections of pre or post conditions.
nil
cljs.user=> cljs.user=> 
``

See command.

## Environment & Version information

*master*

### Lein/Boot version

*Lumo 1.8.0*

### Emacs version

*24.5*

### Operating system

*Ubuntu 16.04*

### LINKED ISSUE #153: Spurious newlines are inserted into output every 1024 characters.
Appears to be related to the newline added here: https://github.com/clojure-emacs/inf-clojure/blob/master/inf-clojure.el#L571.

If that is removed, there are no spurious newlines in the output.

GNU Emacs 26.1 (build 1, x86_64-apple-darwin17.6.0, NS appkit-1561.40 Version 10.13.5 (Build 17F77)) of 2018-06-08

bbatsov:
comint delivers process output in chunks (~1024 bytes), calling preoutput filter functions once per chunk. The old filter had two problems:

1. It prepended `"\n"` to **every** chunk, causing spurious newlines in long output (#153)
2. It checked `this-command`/`last-command` on **every** chunk, which is fragile since the command context can change between chunks (#136)

The fix uses two buffer-local flags to track state across chunks:
- `inf-clojure--filtering-output` — set on the first chunk (where `this-command`/`last-command` is reliable), cleared when a prompt is detected. This makes filtering of subsequent chunks robust.
- `inf-clojure--output-pending-newline` — ensures the newline separator is prepended only once per response.

Also removes the per-chunk `inf-clojure-chomp` call which could strip legitimate newlines from multi-line output.

Fixes #136
Fixes #153

### DIFF
--- a/inf-clojure.el
+++ b/inf-clojure.el
@@ -686,14 +686,41 @@ If `comint-use-prompt-regexp' is nil (the default), \\[comint-insert-input] on
   "Remove subprompts from STRING."
   (replace-regexp-in-string inf-clojure-subprompt "" string))
 
+(defvar-local inf-clojure--filtering-output nil
+  "Non-nil while filtering output from a programmatic inf-clojure command.
+Set on the first output chunk of an inf-clojure command and cleared
+when a prompt is detected, so that multi-chunk responses are filtered
+consistently without relying on `this-command'/`last-command' for
+every chunk.")
+
+(defvar-local inf-clojure--output-pending-newline nil
+  "Non-nil if the next output chunk should be preceded by a newline.
+Used to prepend a single newline before the first chunk of output
+from a programmatic command, separating it from the input.")
+
 (defun inf-clojure-preoutput-filter (str)
-  "Preprocess the output STR from interactive commands."
+  "Preprocess the output STR from interactive commands.
+For output triggered by inf-clojure commands (as opposed to direct
+user input), this removes subprompts and prepends a single newline
+before the first chunk.  Output is tracked across chunks using
+`inf-clojure--filtering-output' so that later chunks are handled
+correctly even if `this-command'/`last-command' has changed."
   (inf-clojure--log-string str "<-RES----")
-  (cond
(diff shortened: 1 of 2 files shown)


candera:
Thank you! This has been bugging me for years.

bbatsov:
Not only you! 😅

